PR #18, opened 2023-12-15
UUID move

Duzos:
moved most things to uuid for performacne (?)


Duzos:
i lied i dont care about your review im merging anyway

Issue #33, opened 2023-12-19
Console Block Entity Rewrite


Loqor:
Currently the ConsoleEnum is stored in the tardis, meaning its global. With that, it would also be easier on the tardis data being stored, and look a lot cleaner than if we kept storing it in the tardisData.

Duzos:
should be safe for merge now : )

Duzos:
issue idk how to fix rn - 
when you break the block the type resets but only on client

Duzos:
to change console right click w command block - needs changing to smth else in future

PR #45, opened 2023-12-23
Redo of console code

Duzos:
look at issue #33 
to change console right click w command block - needs changing to smth else in future
issue idk how to fix rn -
when you break the block the type resets but only on client


Duzos:
misclick

PR #64, opened 2023-12-27
Exterior System Rewrite

Duzos:
Rewrote the entire system to use `ExteriorSchema`s instead of `ExteriorEnum`s as its beter :)
Added a `DoorSchema` which `ExteriorVariantSchema`s can reference for the renderer to use
Other stuff
i kinda forgot
you get the idea

Bugs:
- Any existing worlds will be perma bricked and will insta crash you ( sorry classic )
- Interior doors always use the same texture as the exterior, not too big of a deal for right now
- Probably many, many more.


Duzos:
I did not test exterior changing, shit.

Loqor (approved):
this is so great 10/10 ramble ramble on about how great duzos code is this is very yes so painfully good its incredible craig you should see this suffffff come and see how great this is (please god get his gun out of my face)

PR #168, opened 2024-01-28
Networking Rewrite

Duzos:
Networking is done im pretty sure, just doing some testing of it now to make sure its done.
Should be good with little spam.
Any issues we find now should be menial and easy to fix
Pretty sure this code is disgustingly bad and terrible but it is what it is

what needs doing now is the merge
idk how to merge properly though


Duzos:
Issue - it sends the entire tardis when you leave the interior

Duzos:
What would be even better - partial client tardis data, so when a client joins instead of sending the entire tardis we can just send what they need.
But not yet

Duzos:
This networking thing was done on 1.1.32
some issues might have been fixed, so i need to do this merge to ensure its a me problem
https://github.com/M-D-Team/ait-fabric-1.20.1/compare/1.1.32-1.20.1-beta...1.1.42-1.20.1-beta

PR #271, opened 2024-03-26
Removed src/main/resources/data/ait/structures/interiors/Microsoft Word Project for CIS.pdf and .docx and added an interior image of the default interior but there is no code for it to be added

MaketendoDev:
Removed src/main/resources/data/ait/structures/interiors/Microsoft Word Project for CIS.pdf and .docx and added an interior image of the default interior but there is no code for it to be added


MaketendoDev:
image

Duzos (approved):
LGTM
no code is needed for the image to work iirc 
it’s automatically done

PR #327, opened 2024-06-30
Release Proposal

Duzos:
Release a patch release which fixes some annoying bugs while the next update is in the works.


Loqor:
yeah fuggit why not just call it 1.0.4-hotfix

PR #365, opened 2024-08-21
Hazandra

Duzos:
Teleports you to a linked Tardis' interior if you are companion or higher


Duzos:
todo:
- update texture
- update model
- create recipe

PR #366, opened 2024-08-21
Block Control

Duzos:
Adds blocks which can run controls
Linked by right clicking a control while holding the item

Redstone Control Block runs linked control when powered


Duzos:
imo ready for merge

Issue #305, opened 2024-05-13
Unlockable Dimensions?


Flexico:
I know, I know, the Tardis is always going to be overpowered just by nature, but being able to just pop over to any dimension right out the door might be fixed pretty easily by requiring the player to feed the Tardis a medium-difficulty item from the dimension in question, like netherite scrap for the Nether, chorus fruit for the End, a zanite gem for the Aether, etc. That way you still have to get to the dimension in question via the vanilla route before you can fly there~

Duzos:
could feed it from telepathic circuit, could be cool
but how would modded dimensions work..

Loqor:
We could also just allow you to use a modded item from that dimension ig. could make it data generated, not sure.

miner100910:
Could make it so once the owner or a high loyalty crew member of a TARDIS enters that dimension then the TARDIS is able to go there via some sort of telepathic circuit linking system

DrTheodor:
I side with Masker here.
Some "item from a dimension" is pretty vague, also won't solve the issue for datapack dimensions and situations where the server has a multiworld setup.

DrTheodor:
who the fuck added stale bot

MaketendoDev:
Lololol

Duzos:
> who the fuck added stale bot

😁

PR #381, opened 2024-09-22
Lockable Dimensions

Duzos:
#305 

datapacks can lock a dimension by defining an item needed to unlock it

unlocked by using the telepathic circuit while holding the item

this feature is togglable in the config

### DIFF
--- a/src/main/java/loqor/ait/core/lock/LockedDimension.java
+++ b/src/main/java/loqor/ait/core/lock/LockedDimension.java
@@ -1,0 +1,72 @@
+package loqor.ait.core.lock;
+
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+import com.mojang.serialization.Codec;
+import com.mojang.serialization.JsonOps;
+import com.mojang.serialization.codecs.RecordCodecBuilder;
+import loqor.ait.AITMod;
+import loqor.ait.api.Nameable;
+import loqor.ait.core.tardis.control.impl.DimensionControl;
+import loqor.ait.core.util.WorldUtil;
+import loqor.ait.data.Loyalty;
+import loqor.ait.registry.Registries;
+import loqor.ait.registry.unlockable.Unlockable;
+import net.minecraft.item.ItemStack;
+import net.minecraft.registry.RegistryKey;
+import net.minecraft.registry.RegistryKeys;
+import net.minecraft.text.Text;
+import net.minecraft.util.Identifier;
+import net.minecraft.util.dynamic.Codecs;
+
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicReference;
+
+public record LockedDimension(Identifier dimension, ItemStack stack) implements Unlockable, Nameable {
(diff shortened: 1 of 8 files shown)


src/main/java/loqor/ait/core/lock/LockedDimension.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    @Override
+    public UnlockType unlockType() {
+        return UnlockType.DIMENSION;
+    }
+
+    @Override
+    public Optional<Loyalty> requirement() {
+        return Optional.of(Loyalty.fromLevel(100));
```
DrTheodor, line 60: So wait are they supposed to all unlock on loyalty 100 or whar

src/main/java/loqor/ait/core/lock/LockedDimensionRegistry.java
```diff
@@ -0,0 +1,63 @@
...
+import net.minecraft.item.ItemStack;
+import net.minecraft.item.Items;
+import net.minecraft.world.World;
+import net.minecraft.world.dimension.DimensionTypes;
+
+import loqor.ait.AITMod;
+import loqor.ait.api.TardisEvents;
+import loqor.ait.registry.datapack.SimpleDatapackRegistry;
+
+public class LockedDimensionRegistry extends SimpleDatapackRegistry<LockedDimension> {
```
DrTheodor, line 15: ~~Yeah no what's the point of using `Unlockable` without `UnlockableRegistry`~~

UPD: I've read deeper, this is fine

src/main/java/loqor/ait/core/lock/LockedDimensionRegistry.java
```diff
@@ -0,0 +1,63 @@
...
+    private static final LockedDimensionRegistry instance = new LockedDimensionRegistry();
+
+    // it can go here idc
+    static {
```
DrTheodor, line 19: This should go to `TravelHandler`

src/main/java/loqor/ait/core/tardis/control/impl/TelepathicControl.java
```diff
@@ -178,4 +187,26 @@ public static void locateWithWorldAsync(ServerPlayerEntity player, Tardis tardis
             }
         });
     }
+
+    private static boolean tryUnlockDimension(ServerPlayerEntity player, ItemStack held, ServerTardis tardis) {
```
DrTheodor, line 191: This should go to either `StatsHandler` or the registry

DrTheodor (approved):
LGTM

src/main/java/loqor/ait/core/lock/LockedDimension.java
Duzos: nope and they don't because i didn't add it to the loyalty check
i just put that because it needed a loyalty value

DrTheodor: Thats confusing
Either stop using the unlockable interface here
OR
Rewrite unlockables to have a check method with a generic instead of the requirement

Duzos: those both sound hard

PR #396, opened 2024-10-19
Multidim TARDIS Interior

Duzos:
Title


Duzos:
currently in beta tests, should be completed

Duzos:
This feature is MultiDimensional TARDIS interiors, previously all the interiors were in one minecraft dimension spaced apart. Now each TARDIS gets their own dimension.

Issue #467, opened 2024-10-29
Datapack (de)mat sounds


duzos:
`MatSound` will become stored in the exteriors instead of being hardcoded in `State`
`MatSound` will have a registry so it can be looked up
`MatSound` can be defined by datapacks, a required SoundEvent and optional for the other params

PR #480, opened 2024-10-30
Datapack (de)mat SFX + Animations

Duzos:
#467


Duzos:
i forgot loqor and theo are dead
im merging

PR #518, opened 2024-11-03
Opinions

Duzos:
feat: ItemOpinion, right click TelepathicControl with an item a TARDIS likes for loyalty
api: ItemOpinion can be defined in datapacks, see likes_golden_carrot.json as example


Loqor (changes requested):
Looks good, as long as each TARDIS has a likes list of 8 items that are unique and and can be modified this is perfect.

Duzos:
> Looks good, as long as each TARDIS has a likes list of 8 items that are unique and and can be modified this is perfect.

currently there are 4 items
can you think of 4 other items and the loyalty they should give?

MaketendoDev:
elytra and maybe 20 loyalty?

Loqor:
my thinking was that a random (long) list of items are sorted by rarity, and the rarity dictates a range for their loyalty amount, but that list of items was large and tardises would just have lists of liked items with those loyalty values.

Duzos:
that isnt what i asked 😭 🙏 
if you want that u do it

elytra 20 loyalty sounds good
any more

MaketendoDev:
dragon egg for like 50 or less or more or idk

MaketendoDev:
end crystal for like  15

Issue #564, opened 2024-11-14
Demat Circuit


duzos:
- Required for TARDIS flight
- Slowly loses durability with flight
- Requires #593

PR #601, opened 2024-11-18
Demat Circuit + Structure Backend

Duzos:
#564


Duzos:
todos - model, better names, descriptions, recipe, etc

to test
- fuel up engine
- place down generic subsystem core
- cable to engine
- right click core with demat circuit item
- hear confirmation sfx
- spam zeiton shards to repair demat circuit
- place a 3x3 area of zeiton block beneath block
- hear powerup sfx
- successful takeoff

PR #619, opened 2024-11-30
MultiBlockStructure from NBT

Duzos:
simply place down a generalized subsystem core and then build something around it, then save it as a `.nbt` structure using structure blocks and boom you can load it using `MultiBlockStructure#from`
see `ShieldsCircuit` for example


Duzos:
woops

Duzos:
im merging

PR #729, opened 2024-12-14
Fabricator

Duzos:
Takes in blueprints and inputs and spits out a circuit
Place on top of smithing table
Blueprints are datapack definable

TODO -
- Better sounds
- More recipes


Duzos:
time 2 merge guys

PR #785, opened 2024-12-18
1.21.4 

MaketendoDev:
we'll continuously update this with 1.20.1 (DONT MERGE)


Duzos:
you can convert it to a draft in the top right

MaketendoDev:
1.20.1 and 1.21.4 isn't too big of a difference its just mostly item updates and model changes so any commit should work

MaketendoDev:
Like loqor said theo can do all the work

MaketendoDev:
if we have some kind of item conversion system it should update automatically the item stuff

DrTheodor:
> Like loqor said theo can do all the work 

what.

src/main/java/loqor/ait/core/AITItems.java
```diff
@@ -29,7 +33,7 @@
 import loqor.ait.datagen.datagen_providers.util.NoEnglish;
 
 
-public class AITItems implements ItemRegistryContainer {
+public class AITItems implements AutoRegistryContainer<BlockEntityType<?>> {
```
DrTheodor, line 36: I dont think... thats right..?

DrTheodor (changes requested):
uh
I guess that's not finished

MaketendoDev:
nah it isn't tbf i just pushed what i have so loqor can do some things

MaketendoDev:
its just stale fn until we have a update focused on this

src/main/java/loqor/ait/core/AITItems.java
MaketendoDev: me testing around

PR #872, opened 2025-01-09
Fixed Adaptive Tardis Exterior door 

EnthusiasticMouse:
When The Adaptive Tardis Door is was closed, the disguise would disappear, and when it was open the disguise would appear. (The Opposite should be true)
Described in [This Issue](https://github.com/pavatus/ait/issues/849)

This was caused by a problem where the ChameleonHandler class wasn't getting up to date information on the door state because it was invoked by the USE_DOOR event, which ran before the door was actually updated. 

I've created a new event called DOOR_USED, which is invoked after the door interaction has finished, which fixes the issue.

### DIFF
--- a/src/main/java/loqor/ait/api/TardisEvents.java
+++ b/src/main/java/loqor/ait/api/TardisEvents.java
@@ -168,6 +168,19 @@ public final class TardisEvents {
                 return DoorHandler.InteractionResult.CONTINUE;
             });
 
+    public static final Event<DoorUsed> DOOR_USED = EventFactory.createArrayBacked(DoorUsed.class,
+            callbacks -> (tardis, player) -> {
+                for (DoorUsed callback : callbacks) {
+                    DoorHandler.InteractionResult result = callback.onDoorUsed(tardis, player);
+                    if (result == DoorHandler.InteractionResult.CONTINUE)
+                        continue;
+
+                    return result;
+                }
+
+                return DoorHandler.InteractionResult.CONTINUE;
+            });
+
     public static final Event<EnterTardis> ENTER_TARDIS = EventFactory.createArrayBacked(EnterTardis.class,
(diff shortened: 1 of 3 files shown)


EnthusiasticMouse:
Also apologies in advance If I am not following convention I read the wiki contributing guide but this is my first contribution so might not be perfect :)

src/main/java/loqor/ait/api/TardisEvents.java
```diff
@@ -168,6 +168,19 @@ public final class TardisEvents {
                 return DoorHandler.InteractionResult.CONTINUE;
             });
 
+    public static final Event<DoorUsed> DOOR_USED = EventFactory.createArrayBacked(DoorUsed.class,
```
DrTheodor, line 171: I'd say you should rename the event to AFTER_DOOR_USE and AfterDoorUse to avoid silly mistakes in the future.

src/main/java/loqor/ait/api/TardisEvents.java
```diff
@@ -428,6 +441,10 @@ public interface UseDoor {
         DoorHandler.InteractionResult onUseDoor(Tardis tardis, ServerWorld interior, ServerWorld world, @Nullable ServerPlayerEntity player, @Nullable BlockPos pos);
     }
 
+    @FunctionalInterface
+    public interface DoorUsed {
+        DoorHandler.InteractionResult onDoorUsed(Tardis tardis, ServerPlayerEntity player);
```
DrTheodor, line 446: You should add a DoorHandler.InteractionResult argument to the method, which will be equal to the USE_DOOR's interaction result.

DrTheodor (changes requested):
Other than the stuff i've mentioned everything looks good!
Haven't tested it in-game but this should work.

PR #898, opened 2025-01-19
Library Dependency

Duzos:
This PR
- Adds [sakitus](https://github.com/pavatus/sakitus) as a dependency
- Upgrades gradle to 8.11.1
- Upgrades fabric loom to 1.9


DrTheodor (commented):
legit

PR #1031, opened 2025-03-10
Fix Loyalty Messages

duzos:
@MaketendoDev 
Use this as an example of what to do *instead* of `.append`


MaketendoDev:
We can't say congratulations 😭

PR #1079, opened 2025-03-13
Sonic block and entity scanning

MaketendoDev:
## About the PR
Added a way to scan entity health and blocks to see what tool they need and their blast res plus also shatter glass

## Why / Balance
saturn wanted it and it gives scanning mode more use

## Media
⚔

## Requirements

- [X ] I have read and am following the [Pull Request and Changelog Guidelines](https://amblelabs.github.io/ait-wiki/guidelines).
- [X ] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl:
- add: Added Glass Blocks to be shattered
- add: Added Entity Health Scanning
- add: Added Block Tool Requirement Scanning
-->

### DIFF
--- a/src/main/java/dev/amble/ait/core/item/sonic/ScanningSonicMode.java
+++ b/src/main/java/dev/amble/ait/core/item/sonic/ScanningSonicMode.java
@@ -1,17 +1,5 @@
 package dev.amble.ait.core.item.sonic;
 
-import net.minecraft.entity.player.PlayerEntity;
-import net.minecraft.item.ItemStack;
-import net.minecraft.server.world.ServerWorld;
-import net.minecraft.sound.SoundCategory;
-import net.minecraft.text.Text;
-import net.minecraft.util.*;
-import net.minecraft.util.hit.BlockHitResult;
-import net.minecraft.util.hit.HitResult;
-import net.minecraft.util.math.BlockPos;
-import net.minecraft.util.math.ChunkPos;
-import net.minecraft.world.World;
-
 import dev.amble.ait.core.AITSounds;
 import dev.amble.ait.core.item.SonicItem;
 import dev.amble.ait.core.tardis.Tardis;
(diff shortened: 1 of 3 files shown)


src/main/java/dev/amble/ait/core/item/sonic/ScanningSonicMode.java
```diff
@@ -52,12 +59,46 @@ public boolean startUsing(ItemStack stack, World world, PlayerEntity user, Hand
     public boolean process(ItemStack stack, World world, PlayerEntity user) {
         HitResult hitResult = SonicMode.getHitResult(user);
 
-        if (hitResult instanceof BlockHitResult blockHit)
-            return this.scanRegion(stack, world, user, blockHit.getBlockPos());
+        if (hitResult instanceof BlockHitResult blockHit) {
```
DrTheodor, line 62: you should add a delay like the overload mode does

src/main/java/dev/amble/ait/core/item/sonic/ScanningSonicMode.java
```diff
@@ -52,12 +59,46 @@ public boolean startUsing(ItemStack stack, World world, PlayerEntity user, Hand
...
+            return this.scanBlocks(stack, world, user, blockHit.getBlockPos());
+        } else if (hitResult instanceof EntityHitResult entityHit) {
```
DrTheodor, line 64: no need to chain else's, since you're returning a value

src/main/java/dev/amble/ait/core/item/sonic/ScanningSonicMode.java
```diff
@@ -52,12 +59,46 @@ public boolean startUsing(ItemStack stack, World world, PlayerEntity user, Hand
...
+            return true;
+
+        BlockState state = world.getBlockState(pos);
+        Block block = state.getBlock();
+
+        String blastRes = String.format("%.2f", block.getBlastResistance());
+
+        String toolRequirement = "Any Tool";
+        if (state.isIn(BlockTags.NEEDS_DIAMOND_TOOL)) {
+            toolRequirement = "Diamond Tool";
```
DrTheodor, line 84: you should use translatables here

DrTheodor (changes requested):

duzos (changes requested):
Same as Theo

src/main/java/dev/amble/ait/core/item/sonic/ScanningSonicMode.java
MaketendoDev: ik but how would I return it not clicking on a block or entity?

MaketendoDev: wait
nvm

duzos (approved):
LGTM

Addi3:
what modes activate these functions (for wiki)

PR #1120, opened 2025-03-15
party mode disc

Saturnorsomthing:
i ayent typin allthat
anyway so disc now enables party mode ok cool

### DIFF
--- a/src/main/java/dev/amble/ait/client/renderers/exteriors/ExteriorRenderer.java
+++ b/src/main/java/dev/amble/ait/client/renderers/exteriors/ExteriorRenderer.java
@@ -193,26 +193,7 @@ public class ExteriorRenderer<T extends ExteriorBlockEntity> implements BlockEnt
         float t;
         float s;
 
-        if (tardis.stats().getName() != null && "partytardis".equals(tardis.stats().getName().toLowerCase())) {
-            int m = 25;
-            int n = MinecraftClient.getInstance().player.age / 25 + MinecraftClient.getInstance().player.getId();
-            int o = DyeColor.values().length;
-            int p = n % o;
-            int q = (n + 1) % o;
-            float r = ((float)(MinecraftClient.getInstance().player.age % 25) + h) / 25f;
-            float[] fs = SheepEntity.getRgbColor(DyeColor.byId(p));
-            float[] gs = SheepEntity.getRgbColor(DyeColor.byId(q));
-            s = fs[0] * (1f - r) + gs[0] * r;
-            t = fs[1] * (1f - r) + gs[1] * r;
-            u = fs[2] * (1f - r) + gs[2] * r;
-        } else {
-            float[] hs = new float[]{ 1.0f, 1.0f, 1.0f };
-            s = hs[0];
-            t = hs[1];
-            u = hs[2];
-        }
-
-        if (!tardis.extra().getInsertedDisc().isEmpty()) {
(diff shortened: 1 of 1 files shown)


src/main/java/dev/amble/ait/client/renderers/exteriors/ExteriorRenderer.java
```diff
@@ -212,6 +212,25 @@ private void renderExterior(Profiler profiler, Tardis tardis, T entity, float ti
             u = hs[2];
         }
 
+        if (!tardis.extra().getInsertedDisc().isEmpty()) {
```
duzos, line 215: this is copied from above right
just do
if ((that check) || (the whole other check)) 
instead of duplicating the code

src/main/java/dev/amble/ait/client/renderers/exteriors/ExteriorRenderer.java
```diff
@@ -212,6 +212,25 @@ private void renderExterior(Profiler profiler, Tardis tardis, T entity, float ti
...
+            int q = (n + 1) % o;
+            float r = ((float)(MinecraftClient.getInstance().player.age % 25) + h) / 25f;
+            float[] fs = SheepEntity.getRgbColor(DyeColor.byId(p));
+            float[] gs = SheepEntity.getRgbColor(DyeColor.byId(q));
+            s = fs[0] * (1f - r) + gs[0] * r;
+            t = fs[1] * (1f - r) + gs[1] * r;
+            u = fs[2] * (1f - r) + gs[2] * r;
+        } else {
+            float[] hs = new float[]{ 1.0f, 1.0f, 1.0f };
+            s = hs[0];
```
duzos, line 229: just do 
s = 1
etc

duzos (changes requested):

src/main/java/dev/amble/ait/client/renderers/exteriors/ExteriorRenderer.java
Saturnorsomthing: fair enough will do

src/main/java/dev/amble/ait/client/renderers/exteriors/ExteriorRenderer.java
```diff
@@ -193,7 +193,7 @@ private void renderExterior(Profiler profiler, Tardis tardis, T entity, float ti
         float t;
         float s;
 
-        if (tardis.stats().getName() != null && "partytardis".equals(tardis.stats().getName().toLowerCase())) {
+        if (tardis.stats().getName() != null && "partytardis".equals(tardis.stats().getName().toLowerCase()) || (!tardis.extra().getInsertedDisc().isEmpty())) {
```
duzos, line 196: close, but the brackets matter
change to

if ((tardis.stats().getName() != null && "partytardis".equals(tardis.stats().getName().toLowerCase()) || blah blah disc)

duzos (approved):
there you go
test it out for both party mode and the disc just to be sure and then merge

Saturnorsomthing:
tested it and it works

src/main/java/dev/amble/ait/client/renderers/exteriors/ExteriorRenderer.java
Mansarde: I thought so too at first, but in this case it actually doesn't matter.
Regardless of the brackets, the && is evaluated before the ||
But either way I'd probably still keep the brackets for better maintainability.

PR #1137, opened 2025-03-15
ci: disable spotless failure by default

DrTheodor:
## About the PR
Remove spotless check that fails the build. 

## Why / Balance
Instead of enforcing formatting on EVERY build, apply it every once in a while in a separate PR.

## Requirements


duzos (approved):
LGTM
i think we should add something which automatically runs spotless!

PR #1163, opened 2025-03-16
fix: use client tardis ref in door models

DrTheodor:
## About the PR
This is a follow up PR to the client tardis refs PR. It updates the door models to use the client tardis ref instead of getting it from the block entity.

## Requirements


Loqor (approved):
I can't lie it's largely my fault I keep doing this, I keep saying I'll fix it then never do. I'll stop doing this

PR #1184, opened 2025-03-16
fix: experimental branch thing

Loqor:


duzos (changes requested):
don't do this yet
do this when we're releasing on friday
or we can just make this automated!

PR #1186, opened 2025-03-16
build: if branch is release, then the qualifier is ignored

duzos:
## About the PR
if branch is release, then the qualifier is ignored

## Why / Balance

changed to avoid accidental "beta" release onto platforms

## Technical details

updated `versioning.gradle.` to check if `branch` is release then set `qualifier` to blank

## Media

N/A

## Requirements

## Breaking changes

N/A

**Changelog**

N/A

### DIFF
--- a/versioning.gradle
+++ b/versioning.gradle
@@ -1,16 +1,22 @@
 ext.getCurrentVersion = {
     var patch = System.getenv("GITHUB_RUN_NUMBER")
     var branch = System.getenv("GITHUB_REF")
+    var qualifier = project.mod_version_qualifier
 
     if (branch != null && branch.startsWith("refs/heads/")) {
         branch = branch.substring("refs/heads/".length())
         // replace / with _
         branch = branch.replaceAll("/", "_")
+
+        if (branch == "release") {
+            qualifier = ""
+        }
+
         if (branch == "main" || branch == "release") {
             branch = null
         }
     }
 
     return project.mod_version + (patch != null ? ("." + patch) : "")\
-        + "-" + project.minecraft_version + "-" + project.mod_version_qualifier + (branch != null ? ("-" + branch) : "")
+        + "-" + project.minecraft_version + "-" + qualifier + (branch != null ? ("-" + branch) : "")
 }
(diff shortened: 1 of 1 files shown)


versioning.gradle
```diff
@@ -1,16 +1,22 @@
...
+        }
+
         if (branch == "main" || branch == "release") {
             branch = null
         }
     }
 
     return project.mod_version + (patch != null ? ("." + patch) : "")\
-        + "-" + project.minecraft_version + "-" + project.mod_version_qualifier + (branch != null ? ("-" + branch) : "")
+        + "-" + project.minecraft_version + "-" + qualifier + (branch != null ? ("-" + branch) : "")
```
Mansarde, line 21: If `qualifier` is empty, then there would still be a `-` at the end.
Shouldn't the `"-"` before the `qualifier` also be conditional, similar to how it's being done for the branch?

DrTheodor (approved):
we should probably update this code in future, the amount of ternary operators in one line is just... jeez

Mansarde:
> we should probably update this code in future, the amount of ternary operators in one line is just... jeez

Maybe they could be extracted and put into one block before the return statement:

```java
patch     = patch     != null ? ("." + patch) : "";
qualifier = qualifier != null ? ("-" + qualifier) : "";
branch    = branch    != null ? ("-" + branch) : "";

return project.mod_version + patch + "-" + project.minecraft_version + qualifier + branch;
```

Or, since there already are if-statements for `qualifier` and `branch`, finalize the variables in there and don't use the ternary operators in the return at all (except for `patch`, unless it gets an if-statement too):

```java
...
        if (branch == "release") {
            qualifier = ""
        } else {
            qualifier = "-" + qualifier
        }

        if (branch == "main" || branch == "release") {
            branch = ""
        } else {
            branch = "-" + branch
        }
    }

    return project.mod_version + (patch != null ? ("." + patch) : "")\
        + "-" + project.minecraft_version + qualifier + branch
}
```

To my eyes, putting the ternaries in one block looks neat to the eyes as they're pretty short, whereas in the second example the expanded if-blocks look vertically blown up.
But that's just my personal taste. ^^

duzos:
i like the first one u sent

PR #1187, opened 2025-03-16
feat: sonic scan mode is hand-dependent

### LINKED ISSUE #1106: [FEATURE] Split scanning mode into main-hand vs. off-hand
### Description

After the block/entity scans were added to the sonic's scan mode, we now always have to look into a direction that has >= 17 blocks of air so that a rift/AU scan is done instead of a block/entity-scan.
Outside the TARDIS this will work most of the times, sometimes (in areas with a lot of trees) it's easiest to to look upwards into the sky.
But inside the TARDIS it's more difficult because many console rooms are relatively small and don't have a lot of air-spaces in any one direction that would make it possible to get an Artron-Scan via the sonic.

Usually one would either be looking for a rift *or* want to scan a block/entity.
Therefore it might make sense to split the block/entity scanning from the rift/AU scanning.

With the sonic's "TARDIS" mode we already do such a split of functions by distinguishing between main-hand and off-hand (as well as looking up or down).
Maybe we could do the same for the scanning mode's block/entity vs. rift/AU scan?

**TL;DR**
If in main-hand -> Exclusively do block/entity scans.
If in off-hand -> Exclusively do rift/AU-scans.

PS: Apologies for the "bug" label, but that's because the feature template is still set to that.

N/A

duzos:
## About the PR

main hand to scan blocks/entities
off hand to scan region

closes #1106

## Why / Balance

As the range of block scan is quite large, you have to look in air to scan a region
This is changed by disabling the block scan in off hand mode, allowing region scans to become easier.

## Technical details

add "isMainHand" check to `ScanningSonicMode#process`

## Media

N/A

## Requirements

## Breaking changes

N/A

**Changelog**

:cl:
- feat: sonic scan mode is now hand-dependent (main = blocks/entities, off = region)

### DIFF
--- a/src/main/java/dev/amble/ait/core/item/sonic/ScanningSonicMode.java
+++ b/src/main/java/dev/amble/ait/core/item/sonic/ScanningSonicMode.java
@@ -73,12 +73,16 @@ public class ScanningSonicMode extends SonicMode {
     public boolean process(ItemStack stack, World world, PlayerEntity user) {
         HitResult hitResult = SonicMode.getHitResult(user);
 
-        if (hitResult instanceof BlockHitResult blockHit && !world.getBlockState(blockHit.getBlockPos()).isAir()) {
-            return this.scanBlocks(stack, world, user, blockHit.getBlockPos());
-        }
+        boolean isMainHand = user.getMainHandStack().getItem() == stack.getItem();
 
-        if (hitResult instanceof EntityHitResult entityHit && !(entityHit.getEntity() instanceof RiftEntity)) {
-            return this.scanEntities(stack, world, user, entityHit.getEntity());
+        if (isMainHand) {
+            if (hitResult instanceof BlockHitResult blockHit && !world.getBlockState(blockHit.getBlockPos()).isAir()) {
+                return this.scanBlocks(stack, world, user, blockHit.getBlockPos());
+            }
+
+            if (hitResult instanceof EntityHitResult entityHit && !(entityHit.getEntity() instanceof RiftEntity)) {
+                return this.scanEntities(stack, world, user, entityHit.getEntity());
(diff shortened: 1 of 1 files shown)


src/main/java/dev/amble/ait/core/item/sonic/ScanningSonicMode.java
```diff
@@ -73,12 +73,16 @@ public void tick(ItemStack stack, World world, LivingEntity user, int ticks, int
...
+            if (hitResult instanceof BlockHitResult blockHit && !world.getBlockState(blockHit.getBlockPos()).isAir()) {
+                return this.scanBlocks(stack, world, user, blockHit.getBlockPos());
+            }
 
-        if (hitResult instanceof EntityHitResult entityHit && !(entityHit.getEntity() instanceof RiftEntity)) {
-            return this.scanEntities(stack, world, user, entityHit.getEntity());
+            if (hitResult instanceof EntityHitResult entityHit && !(entityHit.getEntity() instanceof RiftEntity)) {
+                return this.scanEntities(stack, world, user, entityHit.getEntity());
+            }
         }
```
DrTheodor, line 86: im not sure this is intended but
if a player uses the sonic in main hand but the raycast fails then it will scan the region even though the sonic is inthe main hand

duzos: see [my comment](https://github.com/amblelabs/ait/issues/1106#issuecomment-2727623196)

DrTheodor: gotcha

DrTheodor (approved):
LGTM

PR #1188, opened 2025-03-16
feat: subsystem name is displayed when sonic scans the block

### LINKED ISSUE #1179: Show subsystem name when hovering sonic over subsystem core
### Description

Hovering over a subsystem core with the sonic in scanning mode, it would be handy to not only see the durability, but also the name of the installed subsystem.
Reason being that it's a bit difficult to see the subsystem model inside the core block or to recall what model represents which subsystem (especially since some models look similar to each other).

TL;DR
Hover over subsystem block with sonic in scanning mode => Show not only durability _but also_ subsystem name

N/A

duzos:
## About the PR

The subsystem name is now displayed when the block is hovered with a scanning sonic.
A warning about link was made translatable
closes #1179 

## Why / Balance

To improve clarity on what subsystem is inside the block, as its hard to see / remember what item is which.

## Technical details

Use `WorldUtil#fakeTranslate` to translate `Subsystem#getId#name`
"LINK TO ENGINE VIA FLUID LINKS" is translatable from "tardis.message.subsystem.requires_link"

## Media

N/A

## Requirements

## Breaking changes

N/A

**Changelog**

:cl:
- feat: display subsystem name & info on sonic scan


Mansarde:
It looks like the name is touching the previous line, making it overlap the shadow of it:
image
In this example, at first glance it almost looks like the first letter is a `[` and not an `L`.
Maybe a distance of one pixel(?) between the lines would aid legibility?

src/main/java/dev/amble/ait/client/renderers/SonicRendering.java
```diff
@@ -154,7 +156,7 @@ public void renderGui(DrawContext context, float delta) {
         profiler.swap("redstone");
         renderRedstone(context, state, targetPos);
         profiler.swap("durability");
```
DrTheodor, line 158: it doesnt render durability anymore, so the profiler category should be updated too

src/main/java/dev/amble/ait/client/renderers/SonicRendering.java
```diff
@@ -170,21 +172,24 @@ private void renderPower(DrawContext context, BlockPos pos) {
 
         context.drawCenteredTextWithShadow(client.textRenderer, "" + power, getCentreX(), (int) (getMaxY() * 0.4), Colors.WHITE);
     }
-    private void renderDurability(DrawContext context, BlockPos pos) {
+    private void renderSubSystemInfo(DrawContext context, BlockPos pos) {
```
DrTheodor, line 175: line breaks please

src/main/java/dev/amble/ait/client/renderers/SonicRendering.java
```diff
@@ -170,21 +172,24 @@ private void renderPower(DrawContext context, BlockPos pos) {
...
         if (!(client.world.getBlockEntity(pos) instanceof SubSystemBlockEntity be)) return;
 
-        String text = "";
+        Text text = Text.empty();
```
DrTheodor, line 178: move this field down after line 181 (after the return)

src/main/java/dev/amble/ait/client/renderers/SonicRendering.java
```diff
@@ -170,21 +172,24 @@ private void renderPower(DrawContext context, BlockPos pos) {
...
+            text = Text.literal((Math.round(((DurableSubSystem) be.system()).durability())) + " / 1250");
         }
         if (!system.isEnabled() && !(system instanceof EngineSystem)) {
-            text = "LINK TO ENGINE VIA FLUID LINKS";
+            text = Text.translatable("tardis.message.subsystem.requires_link");
         }
 
         context.drawCenteredTextWithShadow(client.textRenderer, text, getCentreX(), (int) (getMaxY() * 0.42), Colors.WHITE);
+
+        text = Text.literal(WorldUtil.fakeTranslate(system.getId().name()));
```
DrTheodor, line 191: its bad to use the `fakeTranslate` method. Because it, well, doesnt translate at all.

make proper translations for the systems and use `Identifier#toTranslationKey`

DrTheodor: its just for better accessibility

DrTheodor: iirc fake translate is used only as a fallback when theres no user friendly (translated) name for a dimension

DrTheodor (changes requested):

src/main/java/dev/amble/ait/client/renderers/SonicRendering.java
DrTheodor: [`toTranslationKey`](https://maven.fabricmc.net/docs/yarn-1.20.1+build.1/net/minecraft/util/Identifier.html#toTranslationKey(java.lang.String)) does exist on Identifiers. We use it in multiple places.

DrTheodor (approved):

Issue #477, opened 2024-10-30
TARDIS duplication on landing


Loqor:
rare, from what i've seen. but definitely still happens

Duzos:
we can't solve this without any kind of footage, logs, etc

Duzos:
![image](https://github.com/user-attachments/assets/4f288aea-b523-4d6f-8978-22eaebb438f4)

Duzos:
new exterior left
old exterior right
taking off again does not fix

Duzos:
old is a real exterior block
perhaps the position was stored wrong / isnt deleted properly?

Duzos:
a hotfix would be exteriors check if they equal the current position and if not delete themselves but idk
@Loqor  thought

DrTheodor:
> a hotfix would be exteriors check if they equal the current position and if not delete themselves but idk

please no
has anybody found a way to replicate the issue consistently?

DrTheodor:
something changes the position and causes `#placeExterior` to run

*Maybe* its trying to materialize mid-air, and then ground search kicks in, then the exterior falls and then you'll have 1 exterior that fell and one that materialized? Should be easy to test this

Duzos:
bump @DrTheodor

DrTheodor:
have you tried to test the stuff ive written a few comments earlier?
If it's not one of those, i really dont know what could be causing the issue

Duzos:
cant repro no more

duzos:
tell a friend to tell a friend
hes baaaack

MaketendoDev:
an Idea is we make it so the exterior block can compare of where its supposed to be and if its not at it then it deletes it self
and then trys to set a exterior block at the pos and it compares when taking off or landing?

duzos:
my assumption is RWF

PR #1226, opened 2025-03-17
chore: delete invalid exteriors on use and on collision

duzos:
## About the PR

Invalid exteriors get deleted

## Why / Balance

patch for #477
RWF causes dupe exteriors, temp patch until 1.3.0

## Technical details

Added a method which checks the expected dimension and position & deletes the exterior if they dont match.
Is ran on use and on collision

## Requirements


src/main/java/dev/amble/ait/core/blockentities/ExteriorBlockEntity.java
MaketendoDev, line 1: can mister @Saturnorsomthing test this so it doesnt delete casual tardis blocks

MaketendoDev (approved):
Apart from testing LGTM

PR #1228, opened 2025-03-17
fix: blueprint tooltip flickering between amounts

duzos:
## About the PR

Blueprint display now shows the range the item count could be in tooltip

## Why / Balance

It was difficult to see the amounts before, as it would flicker between them.

## Technical details

Added `text` method to `Input` for representation.

## Requirements


src/main/java/dev/amble/ait/core/item/blueprint/BlueprintSchema.java
MaketendoDev, line 1: Why are you using Text.of() use Text.literal()

duzos: okay

MaketendoDev (approved):
LGTM

PR #1229, opened 2025-03-17
fix: tardis returning another subsystem core

### LINKED ISSUE #1218: TARDIS returns one too many subsystem cores
### What happened?

Description:
If you install X subsystems and change desktop theme, you will get X + 1 subsystem cores back.

Expected:
You should only get X subsystem cores back.

1. Grow a new TARDIS and only install a console, an engine, as well as one subsystem core.  
(The latter is mainly for life support so you can test this on survival, but if you do this on creative you don't need to install any subsystem at all to test this).
3. Refuel the TARDIS until there is enough Artron energy to change the desktop theme.
4. Change the desktop theme and leave the TARDIS.
5. After re-entering the TARDIS, take a look in the chest next to the entrance.
=> There will be one subsystem core _more_ than subsystems that were installed previously.

### Version

1.2.0

### Build number & branch

187

### Logs & screenshots

N/A

duzos:
## About the PR

closes #1218

## Requirements


MaketendoDev (approved):
Works on my machine

PR #1233, opened 2025-03-17
fix: Engine not breaking its blocks

duzos:
asgyiuml


MaketendoDev (approved):
That actually fixes it LMAO

PR #1251, opened 2025-03-18
Interior Commands (no need to use the TARDIS ID)

Loqor:
## About the PR
I added interior-only commands that can be run from the interior without having to use the tardis id of the tardis you're targeting.

## Why / Balance
Having to constantly remember a tardis' id is super annoying, and most of the time that im using the tardis id specific commands, I'm inside of that tardis anyway.

## Technical details
Just made it so it gets the tardis from the serverWorld, looks like "ServerTardis tardis = ((TardisServerWorld) source.getWorld).getTardis();" inside of new command classes called <TardisCommand>InteriorCommand.

## Requirements

## Breaking changes
NEEDS TO HAVE WARNINGS AND FAIL CASES SO IT DOESN'T CRASH PEOPLE THAT ARE OUTSIDE OF A TARDIS.

:cl:
- Added interior-executable commands.

### DIFF
--- a/src/main/java/dev/amble/ait/AITMod.java
+++ b/src/main/java/dev/amble/ait/AITMod.java
@@ -221,6 +221,7 @@ public class AITMod implements ModInitializer {
         Registry.register(net.minecraft.registry.Registries.FEATURE, CRATER_ID, CRATER);
 
         CommandRegistrationCallback.EVENT.register(((dispatcher, registryAccess, environment) -> {
+            // Normal commands that require a TARDIS ID
             TeleportInteriorCommand.register(dispatcher);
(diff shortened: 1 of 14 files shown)


duzos (changes requested):
Instead you should edit the existing commands so that the tardis ID comes last and is optional

MaketendoDev:
maybe we could also have a version if the exterior is looked at

Mansarde:
> maybe we could also have a version if the exterior is looked at

Maybe similar to the vanilla `/setblock` command (where its first three coordinate parameters are pre-filled based on the block you look at so that you can just press tab to complete the command with them).

PR #1274, opened 2025-03-21
fix: store proper subsystem in the chest on reconfig

### LINKED ISSUE #1265: Subsystems not in recovery chest
### What happened?

**<ins>Description:</ins>**
Currently - when changing interiors - only the Engine as well as the Generalized Subsystem Cores end up in the recovery chest.
_The Subsystems themselves just drop as items near where they were installed. If they fall into the void, they end up at the door a few moments later._
(Sometimes though some of the subsystems end up on the roof of the new console room, depending on their previous location of the old interior.)

**<ins>Expectation:</ins>**
Subsystems should end up in the recovery chest, just like their cores and the engine already do.

</br>

**<ins>Steps to reproduce the behavior:</ins>**
1. Have an engine and a few subsystems installed.
2. Change the interior.
3. After the change, notice how the chest only contains the engine + subsystem cores, while the subsystems will appear at the door after a moment.

<hr>

For a quick orientation, this is the line that gathers the items for the recovery chest: https://github.com/amblelabs/ait/blob/47d13c1d4bf4925861708d622d127d615dbac022/src/main/java/dev/amble/ait/core/tardis/handler/InteriorChangingHandler.java#L188
It calls the `toStacks()` method of each subsystem which currently only adds a "Generalized Subsystem Core", here: https://github.com/amblelabs/ait/blob/47d13c1d4bf4925861708d622d127d615dbac022/src/main/java/dev/amble/ait/core/engine/SubSystem.java#L99

### Version

1.2.0

### Build number & branch

252

### Logs & screenshots

Image

DrTheodor:
## About the PR
This PR fixes #1265.

## Requirements

**Changelog**

:cl:
- fix: store proper subsystem in the chest on reconfig

### DIFF
--- a/src/main/java/dev/amble/ait/core/blocks/EngineBlock.java
+++ b/src/main/java/dev/amble/ait/core/blocks/EngineBlock.java
@@ -99,9 +99,4 @@ public class EngineBlock extends SubSystemBlock implements BlockEntityProvider {
     public BlockRenderType getRenderType(BlockState state) {
         return BlockRenderType.ENTITYBLOCK_ANIMATED;
     }
-
-    @Override
-    protected BlockEntityType<? extends SubSystemBlockEntity> getType() {
-        return AITBlockEntityTypes.ENGINE_BLOCK_ENTITY_TYPE;
-    }
 }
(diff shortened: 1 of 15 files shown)


duzos (approved):
LGTM

duzos:
let's hope they don't still spawn

PR #1280, opened 2025-03-21
fix: the painting no longer move up by one block

### LINKED ISSUE #1259: Gallifrey Falls painting moves up (ONLY ON CLIENT!!!)
### What happened?

Gallifrey Falls painting shifts up, but the effect is purely clientside

1. Grab Gallifrey Falls painting
2. Place painting
3. Wait ~30 seconds
4. Place block in the "empty" space it leaves below it
5. Painting will still break, implying it's client-side only (Loqor confirmed)

### Version

1.2.0

### Build number & branch

_No response_

### Logs & screenshots

N/A

DrTheodor:
## About the PR
This PR fixes #1259 (or should).

## Technical details
Apparently, mc is weird.

## Requirements

**Changelog**

:cl:
- fix: the painting no longer move up by one block

### DIFF
--- a/src/main/java/dev/amble/ait/core/entities/GallifreyFallsPaintingEntity.java
+++ b/src/main/java/dev/amble/ait/core/entities/GallifreyFallsPaintingEntity.java
@@ -2,6 +2,7 @@ package dev.amble.ait.core.entities;
 
 import java.util.Optional;
 
+import net.minecraft.util.math.Vec3d;
 import org.apache.commons.lang3.Validate;
 import org.jetbrains.annotations.Nullable;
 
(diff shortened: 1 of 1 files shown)


Mansarde:
As requested I just tested this on build 285, and it still happens unfortunately.
I.e. After about 50 seconds of placing the painting it still moves up by one block.

Then I re-tried it and also put a block above the painting, hoping it would have no way to go up.
Unfortunately it still moved up (after 55 seconds).
And because it was now partially inside the block I placed above it, the painting's frame darkened.

This probably merits some testing with other paintings, both vanilla and modded paintings, and with various sizes, just to see for which kind of paintings is consistently happens to.
Thoug it could just be another quirk in Minecraft itself and thus a waste of time, especially since it's not yet a craftable painting. 🤷‍♂️

Mansarde:
Ok, tested again, now with build 290.
This time the painting gets placed one block above where I click.
And still it moves one more block up, but it took more than 90 seconds and I left and returned to the TARDIS.

And when I break the block where I originally clicked, the painting (which is now two blocks above that) breaks.

DrTheodor:
what in the world

Mansarde:
Don't break yourself over this though, it's just a painting, not any critical TARDIS function or anything else important. ^^

DrTheodor:
ok last attempt

Mansarde:
Hmm, so far it's staying put, and also exactly where I placed it.
It's been 5 minutes already and I even left the TARDIS inbetween.

Seems like it could actually be fixed now. Congratz! 🍻

DrTheodor:
woooo

DrTheodor:
thank you mr Pan Mansarde for testing

Mansarde:
My pleasure!
And Pan is fine, Mansarde is just my ancient GitHub login before I ever really used it.

Issue #1246, opened 2025-03-18
no id, no linked item


Addi3:
### Description

if a tardis is removed it also deletes any linked items to that tardis so it doesnt spam logs

### Images, links & related resources

NA

### I've checked the feature requests...

- [x] ...and this feature request was not requested before.

DrTheodor:
i mean
its possible to achieve 
but
more of a 1.3.0 goal imo

MaketendoDev:
instead i just made it unlink the item

PR #1284, opened 2025-03-21
fix: linked item spam on unknown tardis

MaketendoDev:
## About the PR
fixed spam console of a linked item to a exterior that doesn't exist

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/src/main/java/dev/amble/ait/api/tardis/link/LinkableItem.java
+++ b/src/main/java/dev/amble/ait/api/tardis/link/LinkableItem.java
@@ -87,6 +87,10 @@ public abstract class LinkableItem extends Item {
         return tardis.getUuid().equals(this.getTardisId(stack));
     }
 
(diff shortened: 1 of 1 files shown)


MaketendoDev:
fixes #1246

DrTheodor:
the reason why we dont just unlink stuff is because the tardis manager can return null in 2 scenarios: 
1. the tardis is missing/corrupted
2. the tardis is not loaded yet

DrTheodor:
and we obv dont want to unlink in the 2nd scenario (the most likely one too)

DrTheodor (changes requested):
read my comment above

MaketendoDev:
can't we make it so you can get all the tardises that exists in the .ait folder

MaketendoDev:
and it could compare to the tardis id that is linked

PR #1285, opened 2025-03-21
fix: turned off boti by defualt

MaketendoDev:
## About the PR
turned off boti (people just wont look at the config like the people they are and will just report that their boti wont work :sob:)

## Requirements

**Changelog**


Saturnorsomthing:
im going to say you really shouldnt do this
since this makes rifts invisible and rifts are a integreal part of the thing
and people are probably to stupid to turn the thing back on

Loqor:
this should only be disabled by default if indium isnt present on amd cards

duzos (approved):
LGTM
however i have concerns that the dependency checker may not be accurate when config is setup

PR #1298, opened 2025-03-23
fix: kill interior drops on interior change end

### LINKED ISSUE #1296: Lingering items after interior change
### What happened?

**<ins>Description:</ins>**
When an interior is changed, then often certain blocks that depend on other blocks (like redstone, paintings, item frames, etc.) pop off when the transformation process replaces their supporting blocks with e.g. air.
These items then either lie around on the floor of the new interior, or fall into the void, reappearing in-front of the interior door.

**<ins>Steps to reproduce the behavior:</ins>**
1. Either set an interior that uses a lot of redstone, paintings, etc. or place a good amount of those items in the current interior.
2. Initiate a change to a new interior, ideally going from a big one to a smaller one (to get a mix of item drops laying on the floor + some falling into the void).
3. Leave the TARDIS, and upon completion of the interior change, re-enter it.
4. Notice how some items will lie around the new interior, while some others will appear at the interior door after a few seconds.

**<ins>Expectation:</ins>**
After changing the interior, no items from the previous interior should remain, including dropped items (i.e. pickup-able items).

**<ins>NOTE:</ins>** This refers exclusively to items from the *previous* interior. Sometimes items from the new interior pop off after re-entering the TARDIS post-change. That should probably be considered a separate issue though.

### Version

1.2.0

### Build number & branch

308

### Logs & screenshots

Image
Image

DrTheodor:
## About the PR
This PR makes it so the interior drops get removed after interior change sequence.
Closes #1296.

## Why / Balance
Previously, all the stuff like flower pots, signs, paintings, item frames, and etc would drop in the interior as entities after the interior reconfig. Now - they dont.

## Requirements

**Changelog**

:cl:
- fix: kill interior drops after interior change

### DIFF
--- a/src/main/java/dev/amble/ait/core/tardis/TardisDesktop.java
+++ b/src/main/java/dev/amble/ait/core/tardis/TardisDesktop.java
@@ -205,15 +205,6 @@ public class TardisDesktop extends TardisComponent {
         this.tardis.door().setLocked(false);
         this.tardis.door().setDeadlocked(false);
         this.tardis.alarm().disable();
-
-        AITMod.LOGGER.info("Removing entities in {}", CORNERS.getBox().expand(0, 300, 0));
-
-        this.tardis.asServer().getInteriorWorld().getEntitiesByClass(
-                ItemEntity.class, CORNERS.getBox().expand(0, 300, 0), i -> true
-        ).forEach(itemEntity -> {
-            AITMod.LOGGER.info("removing {}", itemEntity);
-            itemEntity.discard();;
-        });
     }
 
     public ActionQueue changeInterior(TardisDesktopSchema schema, boolean clear, boolean sendEvent) {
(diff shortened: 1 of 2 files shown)


Mansarde:
It works now, i.e. any item that dropped during the interior change will no longer appear after re-entering the TARDIS.

BUT: It seems that if a painting's or item-frame's supporting block gets replaced with another solid block during the regeneration (i.e. not with air), then those paintings and item frames will survive the regeneration.
Once you re-enter the TARDIS, you will briefly see them hanging on the walls, shortly after which they'll pop off.

So it seems that paintings and item frames themselves (and perhaps similar item types) are not replaced/destroyed during the regeneration.
I assume that's because the interior changing process is focussing on replacing blocks and block-entities?
Not sure if it's possible to include other item types in that process.

MaketendoDev (approved):
Seems 
LGTM

PR #1300, opened 2025-03-23
feat: you can now reference tardises around in commands (^ and ~)

DrTheodor:
## About the PR
This PR makes it so you can use `~` to point to the tardis you're currently in and `^` to point at a linked block entity instead of using a full tardis UUID.

This PR is a (better) alternative to #1251.

## Why / Balance
Previously, you had to find the tardis, then go inside it using a command, then copying its UUID, then doing something else. Now you can just look at it, or just be inside.

## Requirements

## Breaking changes
- `GetInsideCommand` -> `ThisCommand`

**Changelog**

:cl:
- add: now you can use ~ and ^ to point at tardises inside of commands


MaketendoDev (approved):
That's actually pretty smart.

duzos (approved):
LGTM

PR #1308, opened 2025-03-24
various networking & perf improvements

DrTheodor:
## About the PR
This PR changes a few things here and there in the TMs and their networking implementations, as well as removes the chunk forcing.

## Why / Balance
Chunk forcing was removed as it was seen as useless yesterday (23.03.25): the tardis updates the block it stands on when landing, meaning it should fall just fine even without the chunk being force loaded 24/7. 
This PR also changes the ordering of STM operations. Previously, server tardis manager would run networking updates first and then tick all the tardises. This PR changes it to be the other way around.
As a bonus, this PR fixes a memory leak in server tardis manager (dedicated servers unaffected, just the singleplayer).

## Technical details
Turns out, the STM's delta never doesnt get cleared on reset...! This means that a lot of stuff was not being GC'd (server tardis -> interior world -> minecraft server... you know how that goes).

## Requirements

## Breaking changes
- Removed `ForcedChunkUtil`.

**Changelog**

:cl:
- perf: slightly improved networking performance.
- fix: fixed some networking bugs


duzos (approved):
LGTM

PR #1309, opened 2025-03-25
fix: load the chunk before falling

DrTheodor:
## About the PR
This is a follow up to my previous PR that fixes the chunk not being loaded on falling.

## Requirements


src/main/java/dev/amble/ait/core/tardis/handler/travel/TravelHandler.java
```diff
@@ -193,11 +193,10 @@ private ExteriorBlockEntity placeExterior(CachedDirectedGlobalPos globalPos, boo
         if (animate)
             this.runAnimations(exterior);
 
-        BiomeHandler biome = this.tardis.handler(Id.BIOME);
-        biome.update(globalPos);
```
duzos, line 197: rip biome handler

DrTheodor: biome handler updates on its own riiight here https://github.com/amblelabs/ait/blob/main/src/main/java/dev/amble/ait/core/tardis/handler/BiomeHandler.java#L49

Mansarde:
I did some testing (with negative results unfortunately):
- Started a new map with a fresh TARDIS
- Took off with the TARDIS to a new chunk, land types disabled, Y high in the air, manual flight, max speed (with sonic scanning in console)
- After arrival it remained stationary at the target Y coordinates (i.e. didn't fall)
- Exiting the TARDIS makes it fall.
- Took off again to the same coordinates in the air (i.e. in the same still loaded chunk)
- TARDIS still didn't fall, except when exited

Mansarde:
> yeah thats tragic im not sure why that happening and how to fix it except for using safe pos search

Is safe pos search what happens if the TARDIS has land types enabled, to look for a suitable position on the ground?

DrTheodor:
yes

Loqor:
Old af and still hasn't been finished because of course not.
Chunk loading and unloading in Minecraft is a dead pain and I think this needs to be revisited, but this pr is too old and stale to get anything done.

PR #1365, opened 2025-04-01
Refactor how the parser finds identifiers

duzos:
## About the PR

parser now looks for the id inside the animation.json instead of in the name of the animation.json
so instead of the file being named
classic_mat.animation.json

it can just be named
classic.animation.json
with animations "classic_demat" & "classic_mat" inside

## Why / Balance

This just makes more sense, and improves workflow as you can export the animations straight from blockbench ( given theyre properly named ) without needing to worry about renaming the file.

## Technical details

RawLookup is now a `ConcurrentHashMap`
Changed what they store.

## Media

N/A

## Requirements

## Breaking changes

Animation jsons can be named "ANYTHING.animation.json"
the path given inside the "animations" of the .json is the path stored, so rename them appropriately.

**Changelog**


Loqor (approved):
This PR was promised 3000 years ago

PR #1385, opened 2025-04-04
Informative Alarms

### LINKED ISSUE #1350: Informative alarms
### Description

**<ins>Proposal:</ins>**
Extend (or overload) the alarm method with a message String and provide some info with each alarm (except for type B) events, see below).

**<ins>Reasoning:</ins>**
Usually any complex machine will have different alarms for specific events, to make the operator able to distinguish between the causes.
In the mod there is only one alarm, which makes sense for practical reasons.
So instead of having multiple alarms on the console, we could instead provide a message to the player that specifies the reason for the alarm.

<hr>

Currently there are alarms for at least these events (i.e. not necessarily exhaustive list):
- A) Backup power
- A) Engine durability low
- A) Hostile entities in TARDIS
- A) Siege block picked up
- B) Hail mary
- B) Random alarm when TARDIS has a negative alignment
- B) Crash
- B) Self destruct
- B) Desktop reconfiguration

For the B) events, some already print messages themselves, and some are self-explanatory.
For example, if we crash then it should be pretty obvious why the alarm is sounding.
Or if the TARDIS is moody and wants to confuse us with the alarm, then it probably wouldn't tell us the reason (or at least not the *real* reason).

For the A) events one could argue that it's not outright clear what the alarm is for once it commences.
One would need to check Artron levels, engine durability or look for hostiles etc. to figure it out.
And in siege mode a player - not knowing what the alarm is for - doesn't even have a way to to figure it out aside from discovering that the TARDIS can be picked up when in siege mode (with no handbrake on).

<hr>

Having said all of that though, I can also see how leaving the alarm ambiguous increases the need for analysis by the player, which can be exciting gameplay in itself for some.
**Therefore - if the feature gets the green light - it might be best to make "Informative alarms" a toggleable cllient option (in the mods settings).**
That way it caters to different playstyles, as some people prefer clarity while others prefer figuring things out on their own.

N/A

duzos:
## About the PR

Alarms now display a message every cloister toll as their reason & can countdown
closes #1350

## Why / Balance

Allows the user to distinguish between alarms and figure out why they're happening. Makes them more useful.

## Technical details

`Alarm` interface and `Countdown` implementation
`ServerAlarmHandler` has an `ArrayQueue` of alarms to be ran and the current alarm
`Alarm` has a priority, prioritised alarms (eg countdowns) ignore the queue
`AlarmType` enum for generic reusables

## Media

N/A

## Requirements

## Breaking changes

deprecated `ServerAlarmHandler#enabled` as there is logic in the methods now, use `#enable` `#disable` `#isEnabled` instead

**Changelog**

:cl:
feat: informative alarms

### DIFF
--- a/src/main/java/dev/amble/ait/core/tardis/handler/ServerAlarmHandler.java
+++ b/src/main/java/dev/amble/ait/core/tardis/handler/ServerAlarmHandler.java
@@ -1,5 +1,6 @@
 package dev.amble.ait.core.tardis.handler;
 
+import dev.drtheo.queue.api.ActionQueue;
 import net.minecraft.entity.Entity;
 import net.minecraft.entity.TntEntity;
 import net.minecraft.entity.mob.HostileEntity;
@@ -16,6 +17,10 @@ import dev.amble.ait.data.Exclude;
 import dev.amble.ait.data.Loyalty;
 import dev.amble.ait.data.properties.bool.BoolProperty;
 import dev.amble.ait.data.properties.bool.BoolValue;
+import net.minecraft.text.Text;
+import net.minecraft.util.Formatting;
+
+import java.util.*;
 
 // use this as reference for starting other looping sounds on the exterior
 public class ServerAlarmHandler extends KeyedTardisComponent implements TardisTickable {
@@ -25,6 +30,12 @@ public class ServerAlarmHandler extends KeyedTardisComponent implements TardisTi
 
     @Exclude
     private int soundCounter = 0;
+
+    @Exclude
+    private Queue<Alarm> alarms;
+
+    @Exclude
+    private Alarm currentAlarm;
 
     private static final BoolProperty ENABLED = new BoolProperty("enabled", false);
     private static final BoolProperty HOSTILE_PRESENCE = new BoolProperty("hostile_presence", true);
@@ -42,16 +53,44 @@ public class ServerAlarmHandler extends KeyedTardisComponent implements TardisTi
         hostilePresence.of(this, HOSTILE_PRESENCE);
     }
 
+    @Override
+    public void postInit(InitContext ctx) {
+        super.postInit(ctx);
+
+        alarms = new ArrayDeque<>();
+    }
+
+    /**
+     * @deprecated The {@link #enable()} and {@link #disable()} methods have logic in them and should be used instead.
+     * @return The enabled state of the alarm
+     */
+    @Deprecated(forRemoval = true, since = "1.3.0")
     public BoolValue enabled() {
         return enabled;
+    }
+
+    public boolean isEnabled() {
+        return this.enabled.get();
     }
 
     public void enable() {
         this.enabled.set(true);
     }
 
+    public Alarm enable(Alarm cause) {
+        tryStart(cause);
+
+        return cause;
+    }
+
+    public Alarm enable(Text cause) {
+        return enable(() -> Optional.ofNullable(cause));
+    }
+
     public void disable() {
         this.enabled.set(false);
+
+        this.currentAlarm = null;
     }
 
     public BoolValue hostilePresence() {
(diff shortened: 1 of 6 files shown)


src/main/java/dev/amble/ait/core/tardis/handler/travel/CrashableTardisTravel.java
```diff
@@ -101,7 +102,7 @@ default void crash() {
         }
 
         tardis.door().setLocked(true);
-        tardis.alarm().enabled().set(true);
+        tardis.alarm().enable(Text.literal("AHHH IM CRASHINGGG"));
```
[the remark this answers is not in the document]
duzos: HAHA

src/main/java/dev/amble/ait/core/tardis/handler/travel/CrashableTardisTravel.java
```diff
@@ -101,7 +102,7 @@ default void crash() {
         }
 
         tardis.door().setLocked(true);
-        tardis.alarm().enabled().set(true);
+        tardis.alarm().enable(Text.literal("System Alert: TARDIS is experiencing a critical failure."));
```
DrTheodor, line 105: use translatables

src/main/java/dev/amble/ait/core/tardis/handler/ServerAlarmHandler.java
```diff
@@ -42,30 +53,71 @@ public void onLoaded() {
         hostilePresence.of(this, HOSTILE_PRESENCE);
     }
 
+    @Override
+    public void postInit(InitContext ctx) {
+        super.postInit(ctx);
+
+        alarms = new ArrayDeque<>();
```
DrTheodor, line 60: yuck
if this isnt utilized on client, check if youre running on server and only then initialize the field.

src/main/java/dev/amble/ait/core/tardis/handler/ServerAlarmHandler.java
```diff
@@ -42,30 +53,71 @@ public void onLoaded() {
...
+    @Deprecated(forRemoval = true, since = "1.3.0")
     public BoolValue enabled() {
         return enabled;
     }
 
+    public boolean isEnabled() {
+        return this.enabled.get();
+    }
+
     public void enable() {
```
DrTheodor, line 76: you sure you dont want to private that one?

src/main/java/dev/amble/ait/core/tardis/handler/ServerAlarmHandler.java
```diff
@@ -42,30 +53,71 @@ public void onLoaded() {
...
         this.enabled.set(true);
     }
 
+    public Alarm enable(Alarm cause) {
+        tryStart(cause);
+
+        return cause;
+    }
+
+    public Alarm enable(Text cause) {
```
DrTheodor, line 86: imo its better to add alarmtypes enum or something of the like
just so we can have reusable alarm types

DrTheodor (changes requested):

src/main/java/dev/amble/ait/core/tardis/handler/ServerAlarmHandler.java
duzos: i do not

src/main/java/dev/amble/ait/core/tardis/handler/ServerAlarmHandler.java
```diff
@@ -96,6 +154,172 @@ public void tick(MinecraftServer server) {
...
+        AlarmType() {
+            this.translation = "tardis.message.alarm." + this.name().toLowerCase();
+        }
+
+        AlarmType(String translation) {
+            this.translation = translation;
+        }
+
+        @Override
+        public Optional<Text> getAlarmText() {
```
DrTheodor, line 321: why optional

DrTheodor (changes requested):

src/main/java/dev/amble/ait/core/tardis/handler/ServerAlarmHandler.java
DrTheodor: but the `alarmtype` is guaranteed to have an alarm message, right?

MaketendoDev: yeah only the alarm thats triggered by the console does not have an physical message unless "console triggered" counts ig

DrTheodor (approved):

PR #1406, opened 2025-04-08
feat: hopper intake on power converter

MaketendoDev:
## About the PR
hopper intake items on power converter +made a interface for non blockentitys to intake items through hoppers 

## Requirements

**Changelog**

### DIFF
--- a/src/main/java/dev/amble/ait/core/blocks/PowerConverterBlock.java
+++ b/src/main/java/dev/amble/ait/core/blocks/PowerConverterBlock.java
@@ -90,13 +90,6 @@ public class PowerConverterBlock extends DirectionalFluidLinkBlock implements Co
     }
 
     @Override
-    public boolean canAcceptItem(World world, BlockPos pos, ItemStack stack, Direction from) {
-        if (!(world.getBlockEntity(pos) instanceof FluidLinkBlockEntity be)) return false;
-        if (!be.isPowered()) return false;
-        return stack.isIn(AITTags.Items.IS_TARDIS_FUEL);
-    }
-
-    @Override
     public ItemStack insertItem(World world, BlockPos pos, ItemStack stack, Direction from, boolean simulate) {
         if (!(world.getBlockEntity(pos) instanceof FluidLinkBlockEntity be)) return stack;
 
--- a/src/main/java/dev/amble/ait/core/blocks/PowerConverterBlock.java
+++ b/src/main/java/dev/amble/ait/core/blocks/PowerConverterBlock.java
@@ -90,12 +90,15 @@ public class PowerConverterBlock extends DirectionalFluidLinkBlock implements Co
     }
 
     @Override
+    public boolean canAcceptItem(World world, BlockPos pos, ItemStack stack, Direction from) {
+        return stack.isIn(AITTags.Items.IS_TARDIS_FUEL);
+    }
+
+    @Override
     public ItemStack insertItem(World world, BlockPos pos, ItemStack stack, Direction from, boolean simulate) {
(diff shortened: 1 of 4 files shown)


src/main/java/dev/amble/ait/core/blocks/PowerConverterBlock.java
```diff
@@ -87,6 +89,32 @@ public ActionResult onUse(BlockState state, World world, BlockPos pos, PlayerEnt
         return super.onUse(state, world, pos, player, hand, hit);
     }
 
+    @Override
+    public boolean canAcceptItem(World world, BlockPos pos, ItemStack stack, Direction from) {
+        if (!(world.getBlockEntity(pos) instanceof FluidLinkBlockEntity be)) return false;
```
DrTheodor, line 94: imo L94-L95 can be removed from here for performance's sake

src/main/java/dev/amble/ait/mixin/HopperBlockEntityMixin.java
```diff
@@ -0,0 +1,53 @@
...
+        if (!(targetState.getBlock() instanceof ConsumableBlock block)) return;
+
+        for (int slot = 0; slot < inventory.size(); slot++) {
+            ItemStack stack = inventory.getStack(slot);
+            if (stack.isEmpty()) continue;
+
+            if (!block.canAcceptItem(world, targetPos, stack, direction.getOpposite())) continue;
+
+            ItemStack attempt = stack.copyWithCount(1);
+            ItemStack leftover = block.insertItem(world, targetPos, attempt, direction.getOpposite(), false);
```
DrTheodor, line 43: i dont really get it
so theres a hopper's itemstack right?
then you copy it with count 1 and pass into `insertItem`, which then copies it again and decrements the size by 1
and then checks whether the stack is empty, returning an empty global item stack instance or the copied stack otherwise

then this stack gets checked here whether its empty or not, and if it is then it decrements the original stack by 1 and returns...

question: 
1. since you've copied the stack with count 1 doesnt that mean that decrementing it by 1 would result in an empty item stack?
2. if all of this is to add support for stuff that will pass multiple items at the same time, wouldnt it be easier to operate on the original stack directly?

MaketendoDev: for the first question no thats there so then it can simulate a single item transfer, which is what vanilla hoppers do per tick since we're not messing with blockentities with inventorys.

MaketendoDev: second one errrrrrrrrrrrrrrrrrrrrrrrrrrr sure 🤷

src/main/java/dev/amble/ait/core/blocks/PowerConverterBlock.java
```diff
@@ -89,13 +89,6 @@ public ActionResult onUse(BlockState state, World world, BlockPos pos, PlayerEnt
         return super.onUse(state, world, pos, player, hand, hit);
     }
 
-    @Override
-    public boolean canAcceptItem(World world, BlockPos pos, ItemStack stack, Direction from) {
```
DrTheodor: nonono
i say you should let this stay just remove the first 2 if statements that do block entity checks
the stack check should stay

DrTheodor: and then you can just remove the stack check in the `insertItem` method.

MaketendoDev: ohhh okie

DrTheodor (approved):

PR #1418, opened 2025-04-13
feat: editable TARDIS type number!

Saturnorsomthing:
## About the PR

Allows the player to edit the tardis type number via a nametag.
Yes im aware their not a global thingy so not all plaques are the same but thats on purpose to allow for more customization.

## Why / Balance

Is cool

## Technical details

To lazy

## Media

image

## Requirements

## Breaking changes

**Changelog**

🆑 
- Editable TARDIS Type number!

### DIFF
--- a/src/main/java/dev/amble/ait/core/blockentities/PlaqueBlockEntity.java
+++ b/src/main/java/dev/amble/ait/core/blockentities/PlaqueBlockEntity.java
@@ -54,7 +54,9 @@ public class PlaqueBlockEntity extends InteriorLinkableBlockEntity {
     @Override
     public void readNbt(NbtCompound nbt) {
         super.readNbt(nbt);
-        this.customPlaqueText = nbt.getString("CustomPlaqueText");
+        if (nbt.contains("CustomPlaqueText")) {
+            this.customPlaqueText = nbt.getString("CustomPlaqueText");
+        }
         if (this.customPlaqueText == null || this.customPlaqueText.isEmpty()) {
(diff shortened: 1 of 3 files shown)


src/main/java/dev/amble/ait/core/blockentities/PlaqueBlockEntity.java
```diff
@@ -1,14 +1,72 @@
...
+    @Override
+    public void writeNbt(NbtCompound nbt) {
+        super.writeNbt(nbt);
+        nbt.putString("CustomPlaqueText", this.customPlaqueText);
+    }
+
+    @Override
+    public void readNbt(NbtCompound nbt) {
+        super.readNbt(nbt);
+        this.customPlaqueText = nbt.getString("CustomPlaqueText");
```
duzos, line 57: When loading into an existing world, you may get a crash from "getString" without checking if the nbt "contains"
Try it out / Add the check

duzos (changes requested):
LGTM

src/main/java/dev/amble/ait/core/blockentities/PlaqueBlockEntity.java
Saturnorsomthing: https://github.com/user-attachments/assets/f8b53421-7f26-4ee1-8e30-7e617e2e1d81

seems to be fine

DrTheodor: Does it just return null if the value doesn't exist..? Check the javadoc

PR #1451, opened 2025-04-23
Fix save/load waypoint controls behaving inconsistently

Mansarde:
## About the PR
The save/load controls, as well as the waypoint ejection from the console port behave, inconcistently.

The way it works before this PR is:
As soon as a cartridge is put into the port, it is being written to.
This means that if you eject (or try to load from) the cartridge without ever having pressed the save control, it will not be empty but filled with the current coordinates.
Also, once the save control is pushed last (between it and the load control), then putting in cartridges at any point afterwards will (over-)write and eject them immediately.

<details>
<summary>Steps to reproduce</summary>

1. Click the load control (just to get into the correct state).
2. Put an *empty* cartridge into the console port.
3. Punch (i.e. left-click) the console port.
=> Notice how the ejected cartridge contains coordinates even though the save control wasn't pushed.
4. Put another *empty* cartridge into the console port.
5. Press the save control.
=> A cartridge with coordinates gets ejected, as expected.
6. Travel with the TARDIS to different coordinates.
7. Put the cartridge from step 5 into the console port.
=> Notice how, before you can even press the load button, the cartridge gets automatically overwritten with the new coordinates and ejected.
***Until you push the load control***, every time you put another cartridge into the console port (empty or not), that cartridge gets ejected immediately with coordinates on it, without you having to push the save control.
</details>

This PR changes the save/load behavior, so that:
- cartridges are only being written to when the save control is pressed
- empty cartridges can be ejected empty
- cartridges are only ejected when save/load control was pushed or port was punched

<details>
<summary>How the steps work after the PR</summary>

1. Put an *empty* cartride into the console port.
2. Punch (i.e. left-click) the console port.
=> Notice how the ejected cartridge remains empty (and even keeps its custom name if it was given one).
3. Put an *empty* cartridge into the console port again.
4. Press the save control.
=> A cartridge with coordinates gets ejected, as expected.
5. Put yet another *empty* cartridge into the console port.
=> Notice how the cartridge isn't automatically ejected.
6. Punch the console port.
=> Notice how the ejected cartridge is still empty.
7. Put a cartridge with coordinates into the console port.
8. Punch the console port.
=> Notice how the coordinates of the ejected cartridge remain unchanged.
Only explicitly pressing the save control will (over-)write to the cartridge!
</details>

In addition, the save control did not make any sound, because its assigned sound-file was missing.
And finally the save control was not able to be part of a flight sequence, contrary to the load control.

## Why / Balance
Having the cartridge be written to without the save control being pushed is counter-intuitive.
This unexpected behavior could easily lead to unintentionally overwriting a cartridge that one intended to load from (as demonstrated in step 7 of the "Steps to reproduce" in the about section above).
Therefore this PR aims to make the behavior more consistent and intuitive by only having the cartridge being acted upon when one of the waypoint controls is pushed.

## Technical details
I modified the `ConsolePortControl`, so that when a cartridge item is put into the port, the current position is no longer saved at that moment.
This required me to add a null-check in `SetWaypointControl` for when the load control is pushed, so that the destination would only be set if there were actually coordinates on the cartridge.
And a similar null-check was added to `WaypointItem` for when the cartridge gets ejected, so that the cartridge item that gets created in the world can be empty (i.e. without coordinates).

To fix the save control not being able to be part of a flight sequence, I simply added a call to super, analogous to how the load control already does it.
To fix the missing sound, I copied the sound-file of the load control for re-use with the save control.

## Requirements

**Changelog**
:cl:
- fix: save waypoint control has no sound
- fix: waypoint cartridge ejected immediately without interaction in some cases
- fix: save/load waypoint controls behave inconsistently and can lead to unexpected outcome
- fix: save control not able to be part of flight sequence


src/main/resources/assets/ait/sounds/controls/waypoint_mark.ogg
DrTheodor, line 1: ... but where did the `mark_waypoint.ogg` go..?

src/main/java/dev/amble/ait/core/tardis/handler/WaypointHandler.java
```diff
@@ -44,16 +44,14 @@ private void clearCartridge() {
...
     public Optional<Waypoint> set(Waypoint var, BlockPos console, boolean spawnItem) {
-        Optional<Waypoint> prev = Optional.ofNullable(this.current);
-
         if (spawnItem && this.current != null)
-            this.spawnItem(console, prev.get());
+            this.spawnItem(console, var);
 
         this.current = var;
-        return prev;
+        return Optional.ofNullable(this.current);
```
DrTheodor, line 54: This violates the contract of the method.
This should return the previous waypoint, but with your changes it returns the current one.
Either change the documentation and all the usages of this method to be aware of the change and add it to the "Breaking changes" or make it return the previous waypoint, as it was before

DrTheodor (changes requested):

src/main/resources/assets/ait/sounds/controls/waypoint_mark.ogg
Mansarde: There never was a `mark_waypoint.ogg`, that's why I added `waypoint_mark.ogg` as a copy of `waypoint_set.ogg`.

src/main/java/dev/amble/ait/core/tardis/handler/WaypointHandler.java
```diff
@@ -44,13 +44,12 @@ private void clearCartridge() {
...
      *
-     * @return The optional of the previous waypoiint
+     * @return The optional of the previous waypoint
      */
     public Optional<Waypoint> set(Waypoint var, BlockPos console, boolean spawnItem) {
         Optional<Waypoint> prev = Optional.ofNullable(this.current);
-
         if (spawnItem && this.current != null)
-            this.spawnItem(console, prev.get());
+            this.spawnItem(console, var);
```
DrTheodor, line 52: wont that drop the waypoint cartridge you're trying to set instead of the one inside..?

Mansarde: The only time that the `set` method will spawn the cartridge item is when the save control is pressed, because it is the only control that passes `true` to it.
When that happens, the waypoint given to the `set` method contains the current coordinates which we want to save to the cartridge that is to be ejected, so we want `var` and not `prev` in this case.

The other two controls (i.e. load control and console port) call `spawnItem` *directly* and not via the `set` method.

That the `set` method contains a spawnItem instruction is a holdover from how it worked before, where cartridges got written to the moment they were put into the slot.
I could remove the spawning in `set` altogether to make the save control consistent with the other two controls. That would also clean up the then redundant boolean argument.

I'll prepare a new commit for an easy comparison and report back.

Mansarde: Ok, pushed a new commit where I move the `spawnItem` holdover completely out of the `set` method.
Now the save control first sets the waypoint and then spawns/ejects the cartridge.
That way we also don't need the boolean parameter for `set` any longer, so I removed that as well.

Mansarde: @DrTheodor Not sure if you got a ping, but I pushed a new commit a while ago to make your requested changes. :)

Mansarde:
Ok, scratch that.
Loqor just fixed it himself with 881b841, and way better. :D

PR #1453, opened 2025-04-23
fix handlers being null sometimes due to multithreading

DrTheodor:
## About the PR
This PR fixes a long standing bug in AIT (since, like, 1.0.4?). The bug that was randomly making handlers go null for no reason is now fixed!

## Technical details
The enum map used to store the handlers is using an array to store the handlers. Rarely, when the networking thread set the object but before it would put the new value in, the rendering thread would intervene and try to get the value, resulting in a null. This PR fixes it by using `synchronized` on a lock for array access.

## Requirements

**Changelog**

🆑
- fix: no more random crashes due to multithreading


Loqor (approved):
LGTM

PR #1459, opened 2025-04-23
fix(manager): crash when loading & file manager locked

### LINKED ISSUE #1443: Loading TARDIS can be null!
### What happened?

This happened!

Link a key (also have loyalty high enough to snap)
turn power off
(maybe) stand close to interior door
press snap keybind while holding said linked key

### Version

1.2.0

### Build number & branch

1.2.0.2805-1.20.1

### Logs & screenshots

server log (pastebin is being a bitch about it so here is the file) > 

[2025-04-21-5.log](https://github.com/user-attachments/files/19838713/2025-04-21-5.log)

duzos:
## About the PR

Return an exception if file manager is locked while trying to load tardis
closes #1443

## Why / Balance

Causes an edge-case crash + makes more sense / continuity

## Technical details

In `loadTardis` 
```
        if (this.fileManager.isLocked())
            return Either.right(new Exception("TardisFileManager is locked!"));
```

## Requirements

### DIFF
--- a/src/main/java/dev/amble/ait/core/tardis/manager/old/DeprecatedServerTardisManager.java
+++ b/src/main/java/dev/amble/ait/core/tardis/manager/old/DeprecatedServerTardisManager.java
@@ -123,8 +123,13 @@ public abstract class DeprecatedServerTardisManager extends TardisManager<Server
 
     @Override
     public void loadTardis(MinecraftServer server, UUID uuid, @Nullable Consumer<ServerTardis> consumer) {
-        if (consumer != null)
-            this.loadTardis(server, uuid).ifLeft(consumer);
+        if (consumer == null) return;
+
+        Either<ServerTardis,Exception> either = this.loadTardis(server, uuid);
+
+        if (either == null) return;
+
+        either.ifLeft(consumer);
     }
 
     @Override


src/main/java/dev/amble/ait/core/tardis/manager/old/DeprecatedServerTardisManager.java
```diff
@@ -152,9 +152,9 @@ public void forEach(Consumer<ServerTardis> consumer) {
         this.lookup.forEach((uuid, either) -> either.ifLeft(consumer));
     }
 
-    @Nullable private Either<ServerTardis, Exception> loadTardis(MinecraftServer server, UUID uuid) {
+    private Either<ServerTardis, Exception> loadTardis(MinecraftServer server, UUID uuid) {
         if (this.fileManager.isLocked())
-            return null;
+            return Either.right(new Exception("TardisFileManager is locked!"));
```
DrTheodor, line 157: uhhhhhhhhhhhhhhhhhhhhhhh
no.
thats not how it should be done

this method should return null
but it should be handled properly wherever it gets called
its not private for no reason

DrTheodor: if it puts an exception inside the map it wont ever try to load that tardis
so if, for example, the TM is locked and then some BlockEntity tries to load its tardis instance it will silently fail which is expected
but when it will try again, it will check the map, find an exception and go on with its day thinking that there was no tardis bound to begin with

DrTheodor (approved):
LGTM
but doesnt the original issue report state that the `loadTardis` method was called from `getTardis`..?

duzos:
yes, because that was null

PR #1470, opened 2025-04-24
Fix disabling of experimental message for map backup-prompt

Mansarde:
## About the PR
Currently when the "Show Experimental Warning" is set to "no", then the experimental-warning will only be disabled for the map _creation_ dialog. But it is also supposed to be disabled when opening existing maps, which it fails to do though.

This PR fixes this, so that the warning messages are not just disabled when creating a new map, but also when opening existing AIT maps.

## Why / Balance
Because it was supposed to work, as evidenced by the existence of the mixin made for that purpose ([WorldOpenFlowsMixin](https://github.com/amblelabs/ait/blob/main/src/main/java/dev/amble/ait/mixin/client/experimental_screen/WorldOpenFlowsMixin.java)).

## Technical details
If the `SHOW_EXPERIMENTAL_WARNING` setting is set to `false`, then the `confirmWorldCreation` method of the [WorldOpenFlowsMixin](https://github.com/amblelabs/ait/blob/main/src/main/java/dev/amble/ait/mixin/client/experimental_screen/WorldOpenFlowsMixin.java#L19) is supposed to prevent the MC backup-screen dialog when opening an existing world, but fails to do so.
Aside from that, it would also get called when creating a new map, which is redundant though because of the [CreateWorldScreenMixin](https://github.com/amblelabs/ait/blob/main/src/main/java/dev/amble/ait/mixin/client/experimental_screen/CreateWorldScreenMixin.java) already taking care of that (successfully).

So instead I hook into an earlier method of `IntegratedServerLoader`, which looks like this:
```java
public void start(Screen parent, String levelName) {
    this.start(parent, levelName, false, true);
}
```

The last boolean parameter determines if a backup screen is shown, so I simply replace this call and set that parameter to `false`.
This successfully prevents the backup screen from showing (if AIT's `SHOW_EXPERIMENTAL_WARNING` is disabled).

## Requirements

**Changelog**
:cl:
- fix: disabled experimental messages still shown during opening of existing AIT maps

### DIFF
--- a/src/main/java/dev/amble/ait/mixin/client/experimental_screen/WorldOpenFlowsMixin.java
+++ b/src/main/java/dev/amble/ait/mixin/client/experimental_screen/WorldOpenFlowsMixin.java
@@ -1,25 +1,25 @@
 package dev.amble.ait.mixin.client.experimental_screen;
 
-import com.mojang.serialization.Lifecycle;
+import net.minecraft.client.gui.screen.Screen;
 import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Inject;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 
-import net.minecraft.client.MinecraftClient;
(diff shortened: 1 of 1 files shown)


duzos (approved):
LGTM

PR #1474, opened 2025-04-24
api(travel): temporary (de)mat anims for single flights

### LINKED ISSUE #983: Some easy way to set the travel sounds
A way to set the remat, demat and flight sound easly via code.
As a example you could do
if tardis.autopilot(true);
   tardis.travel().forSound(FLIGHT).setSound(STABLALIZED)

(really bad example i know)
so now whenever autopilot is true the sound is forced to stabalized regardless of what the user does.
however an alternative example would be putting this in phaser so you could do 

tardis.travel().forSound(DEMAT).setSound(PHASINGDEMAT)
tardis.travel().forSound(FLIGHT).setSound(UNSTABLE)
tardis.travel().forSound(MAT).setSound(PHASINGREMAT)

however since thats 3 lines an alternative to that would be creating "sound packages" (if that makes sense)
eg rather then having 3 lines of code a sound is packaged into 1 file, so as a example we have

a zip called "PHASING", that said zip then contains demat, flight and remat
so then rather then having 3 lines you could do
tardis.travel().forSound(ALL).setPackage(PHASING)
this then sets all 3 sounds to the phaser ones.

however however, forcing the sound to change each time seems a bit annoying since the user then has to set it back to what they want, so heres another thing
tardis.travel().forSound(MAT).tempSound(MASTERMAT)
this dosent set the sound but just makes the sound play that 1 time, so if phaser went off the sound would be "queued" to play the next time the TARDIS lands maby.

duzos:
## About the PR

You can now pass in two animations when calling demat, which will only be used once.
closes #983

## Why / Balance

For features like engine phasing, where you want a custom demat but dont want to replace the player chosen one permamently 

## Technical details

Added `TardisAnimation` to `dematerialize` and `forceDemat` and utilised `ActionQueue` to reset the values after the next state is reached.

## Requirements

## Breaking changes

`dematerialize(TravelSound)` -> `dematerialize(TardisAnimation, TardisAnimation)`
`forceDemat(TravelSound)` -> `forceDemat(TardisAnimation, TardisAnimation)`

### DIFF
--- a/src/main/java/dev/amble/ait/core/engine/impl/EngineSystem.java
+++ b/src/main/java/dev/amble/ait/core/engine/impl/EngineSystem.java
@@ -5,6 +5,7 @@ import java.util.List;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
+import dev.amble.ait.core.tardis.animation.v2.datapack.TardisAnimationRegistry;
 import dev.amble.lib.util.ServerLifecycleHooks;
 import org.joml.Vector3f;
 
@@ -174,9 +175,8 @@ public class EngineSystem extends DurableSubSystem {
             return new Phaser(
                     (phaser) -> {
                         ServerTardis tdis = sTardis.asServer();
-                        TardisUtil.sendMessageToInterior(tdis, Text.translatable("tardis.message.engine.phasing").formatted(Formatting.RED));
                         TardisUtil.sendMessageToLinked(tdis, Text.translatable("tardis.message.engine.phasing").formatted(Formatting.RED));
-                        tdis.alarm().enabled().set(true);
+                        tdis.alarm().enable(Text.translatable("tardis.message.engine.phasing").formatted(Formatting.RED));
                         tdis.getDesktop().playSoundAtEveryConsole(AITSounds.HOP_DEMAT);
                         tdis.getExterior().playSound(AITSounds.HOP_DEMAT);
                         sTardis.subsystems().demat().removeDurability(5);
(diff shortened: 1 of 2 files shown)


src/main/java/dev/amble/ait/core/engine/impl/EngineSystem.java
```diff
@@ -189,7 +189,7 @@ public static Phaser create(EngineSystem system) {
                                 sTardis.travel().speed(500);
                                 sTardis.getDesktop().playSoundAtEveryConsole(AITSounds.UNSTABLE_FLIGHT_LOOP);
                                 sTardis.getExterior().playSound(AITSounds.UNSTABLE_FLIGHT_LOOP);
-                                sTardis.travel().forceDemat(TravelSoundRegistry.PHASING_DEMAT);
+                                sTardis.travel().forceDemat(TardisAnimationRegistry.getInstance().instantiate(AITMod.id("crumple")), TardisAnimationRegistry.getInstance().instantiate(AITMod.id("ghost_mat"))); // TODO - make phasing anims!! - duzo
```
DrTheodor, line 192: Move the "crumple" id to the tardis animation regisyry

DrTheodor (dismissed):

src/main/java/dev/amble/ait/core/engine/impl/EngineSystem.java
duzos: no cus its a json file

Loqor (approved):
LGTM

PR #1478, opened 2025-04-24
Fix disabling of experimental message for map backup-prompt (take two)

### LINKED ISSUE #1477: Previous PR to fix backup warning causes crash
### What happened?

PR #1470 works when testing via `gradlew runClient`, but leads to a crash with the built JAR.
I've already made a PR to revert this temporarily (#1476) which has already been approved, so that the beta branch is usable again.

This has been first reported by @Addi3 on Discord and I've already reproduced it.
I've self-assigned the issue and will investigate how to fix it, keeping this issue for tracking purposes.

Steps to reproduce the behavior (in :
1. Start the game, try to open a new map.
2. See error

### Version

1.2.0

### Build number & branch

689

### Logs & screenshots

The pertinent part:

<details>

```
Caused by: org.spongepowered.asm.mixin.injection.throwables.InjectionError: Critical injection failure: Callback method skipBackupScreen(Lnet/minecraft/class_437;Ljava/lang/String;Lorg/spongepowered/asm/mixin/injection/callback/CallbackInfo;)V in ait.mixins.json:client.experimental_screen.WorldOpenFlowsMixin from mod ait failed injection check, (0/1) succeeded. Scanned 0 target(s). Using refmap ait-refmap.json
    at org.spongepowered.asm.mixin.injection.struct.InjectionInfo.postInject(InjectionInfo.java:531)
    at org.spongepowered.asm.mixin.transformer.MixinTargetContext.applyInjections(MixinTargetContext.java:1490)
    at org.spongepowered.asm.mixin.transformer.MixinApplicatorStandard.applyInjections(MixinApplicatorStandard.java:752)
    at org.spongepowered.asm.mixin.transformer.MixinApplicatorStandard.applyMixin(MixinApplicatorStandard.java:330)
    at org.spongepowered.asm.mixin.transformer.MixinApplicatorStandard.apply(MixinApplicatorStandard.java:246)
    at org.spongepowered.asm.mixin.transformer.TargetClassContext.apply(TargetClassContext.java:437)
    at org.spongepowered.asm.mixin.transformer.TargetClassContext.applyMixins(TargetClassContext.java:418)
    at org.spongepowered.asm.mixin.transformer.MixinProcessor.applyMixins(MixinProcessor.java:363)
    ... 36 more
```
</details>

Mansarde:
## About the PR
Currently when the "Show Experimental Warning" is set to "no", then the experimental-warning will only be disabled for the map _creation_ dialog. But it is also supposed to be disabled when opening existing maps, which it fails to do though.

This PR fixes this, so that the warning messages are not just disabled when creating a new map, but also when opening existing AIT maps.

</BR>

**<ins>NOTE:</ins>** This is the second attempt of the PR.
The previous attempt (#1470) was rolled back (#1476) due to no mappings being generated for the JAR build.
But this time the mappings are implicity generated for the mixin (i.e. removed `remap = false`), not leading to a crash any longer. :)
Therefore, fixes #1477

## Why / Balance
Because it was supposed to work, as evidenced by the existence of the mixin made for that purpose ([WorldOpenFlowsMixin](https://github.com/amblelabs/ait/blob/main/src/main/java/dev/amble/ait/mixin/client/experimental_screen/WorldOpenFlowsMixin.java)).

## Technical details
If the `SHOW_EXPERIMENTAL_WARNING` setting is set to `false`, then the `confirmWorldCreation` method of the [WorldOpenFlowsMixin](https://github.com/amblelabs/ait/blob/main/src/main/java/dev/amble/ait/mixin/client/experimental_screen/WorldOpenFlowsMixin.java#L19) is supposed to prevent the MC backup-screen dialog when opening an existing world, but fails to do so.
Aside from that, it would also get called when creating a new map, which is redundant though because of the [CreateWorldScreenMixin](https://github.com/amblelabs/ait/blob/main/src/main/java/dev/amble/ait/mixin/client/experimental_screen/CreateWorldScreenMixin.java) already taking care of that (successfully).

So instead I hook into an earlier method of `IntegratedServerLoader`, which looks like this:
```java
public void start(Screen parent, String levelName) {
    this.start(parent, levelName, false, true);
}
```

The last boolean parameter determines if a backup screen is shown, so I simply replace this call and set that parameter to `false`.
This successfully prevents the backup screen from showing (if AIT's `SHOW_EXPERIMENTAL_WARNING` is disabled).

## Requirements

**Changelog**
:cl:
- fix: disabled experimental messages still shown during opening of existing AIT maps

### DIFF
--- a/src/main/java/dev/amble/ait/mixin/client/experimental_screen/WorldOpenFlowsMixin.java
+++ b/src/main/java/dev/amble/ait/mixin/client/experimental_screen/WorldOpenFlowsMixin.java
@@ -1,25 +1,25 @@
 package dev.amble.ait.mixin.client.experimental_screen;
 
-import com.mojang.serialization.Lifecycle;
+import net.minecraft.client.gui.screen.Screen;
 import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Inject;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 
-import net.minecraft.client.MinecraftClient;
-import net.minecraft.client.gui.screen.world.CreateWorldScreen;
 import net.minecraft.server.integrated.IntegratedServerLoader;
 
 import dev.amble.ait.AITMod;
 
 @Mixin(value = IntegratedServerLoader.class)
-public class WorldOpenFlowsMixin {
+public abstract class WorldOpenFlowsMixin {
 
-    @Inject(method = "tryLoad", at = @At(value = "INVOKE_ASSIGN", target = "Lcom/mojang/serialization/Lifecycle;experimental()Lcom/mojang/serialization/Lifecycle;", remap = false), cancellable = true)
-    private static void confirmWorldCreation(MinecraftClient client, CreateWorldScreen parent, Lifecycle lifecycle,
-            Runnable loader, boolean bypassWarnings, CallbackInfo ci) {
+    @Shadow protected abstract void start(Screen parent, String levelName, boolean safeMode, boolean canShowBackupPrompt);
+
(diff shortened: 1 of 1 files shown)


duzos (approved):
LGTM

PR #1485, opened 2025-04-25
Feat: Max range of 256 blocks for sonic summon

Mansarde:
## About the PR
This adds a max range of 256 blocks to the sonic summon.

## Why / Balance
@Ouroboros-Games menioned in Discord that this limitation was in before the sonic rewrite and it was meant to give the Stattenheim Remote more reason to exist.

## Technical details
I saw that the old sonic code used `isNearTardis()` and this method still exists in [ExteriorAnimation](https://github.com/amblelabs/ait/blob/main/src/main/java/dev/amble/ait/core/tardis/animation/ExteriorAnimation.java#L93).
But because that class is marked as deprecated, I copied its code (which is all self-contained) to the `TardisUtil` class and make use of it from there.

So in the `TardisSonicMode` class I simply add a `isNearTardis()` check as a guard clause to determine if the player is within 256 blocks of the TARDIS.
If they are, the code just continues as before.
If they're not, then I display a (translatable) message to let the user know and return false.

## Requirements

**Changelog**
:cl:
- add: max range of 256 blocks for sonic summon

### DIFF
--- a/src/main/java/dev/amble/ait/core/tardis/util/TardisUtil.java
+++ b/src/main/java/dev/amble/ait/core/tardis/util/TardisUtil.java
@@ -465,4 +465,14 @@ public class TardisUtil {
 
         return Optional.ofNullable(nearestPlayer);
     }
+
+    public static boolean isNearTardis(PlayerEntity player, Tardis tardis, double radius) {
+        return radius >= distanceFromTardis(player, tardis);
+    }
+
+    public static double distanceFromTardis(PlayerEntity player, Tardis tardis) {
+        BlockPos pPos = player.getBlockPos();
+        BlockPos tPos = tardis.travel().position().getPos();
+        return Math.sqrt(tPos.getSquaredDistance(pPos));
+    }
 }
--- a/src/main/java/dev/amble/ait/core/item/sonic/TardisSonicMode.java
+++ b/src/main/java/dev/amble/ait/core/item/sonic/TardisSonicMode.java
@@ -99,9 +99,11 @@ public class TardisSonicMode extends SonicMode {
             return false;
         }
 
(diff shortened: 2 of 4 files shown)


src/main/java/dev/amble/ait/core/tardis/util/TardisUtil.java
```diff
@@ -465,4 +465,14 @@ public static Optional<ServerPlayerEntity> findNearestPlayer(CachedDirectedGloba
...
     }
+
+    public static boolean isNearTardis(PlayerEntity player, Tardis tardis, double radius) {
+        return radius >= distanceFromTardis(player, tardis);
+    }
+
+    public static double distanceFromTardis(PlayerEntity player, Tardis tardis) {
+        BlockPos pPos = player.getBlockPos();
+        BlockPos tPos = tardis.travel().position().getPos();
+        return Math.sqrt(tPos.getSquaredDistance(pPos));
```
duzos, line 476: Ensure they are in the same dimension as well.

duzos (changes requested):

src/main/java/dev/amble/ait/core/tardis/util/TardisUtil.java
Mansarde: K, I'll look into how to best do that. Dimension stuff is new to me.
Will see how it's done in other places in the mod.

Mansarde: > Dimension check should be done in sonic code

Ok thx!

duzos (approved):
LGTM

Mansarde:
> LGTM

Thx, but I can't merge, you'll have to do that. :)

PR #1507, opened 2025-04-28
make control entities regular entities

DrTheodor:
## About the PR
This PR makes it so the control entities are now regular entities.

## Why / Balance
Before, they were living entities which were adding quite a bit of overhead.

## Technical details
This also removes a hack-fix mixin.

## Requirements

## Breaking changes

**Changelog**

🆑
- perf: control entities are no longer living entities


Loqor (changes requested):
Please make the entities never despawn and/or store their data somewhere else so we don't constantly have to relink them and so their durability doesn't constantly get reset.

duzos (approved):
LGTM

PR #1509, opened 2025-04-29
Lazy tardis ticking

DrTheodor:
## About the PR
This PR adds lazy tardis ticking.

## Why / Balance
Previously, the tardises would tick in any condition, even if they are not loaded outside or inside. Now they don't!

## Technical details
Besides that, the PR fixes AIT to lazily load the world instead of getting it loaded right away. Unfortunately, as of the currently used version of multidim this won't actually guarantee lazy world loading.

## Requirements

## Breaking changes
- `TardisDesktop#getInteriorWorld` -> `TardisDesktop#worldRef`

**Changelog**

:cl:
- perf: improved ticking large amounts of tardises

### DIFF
--- a/src/main/java/dev/amble/ait/compat/gravity/GravityHandler.java
+++ b/src/main/java/dev/amble/ait/compat/gravity/GravityHandler.java
@@ -1,19 +1,19 @@
 package dev.amble.ait.compat.gravity;
 
-import java.util.List;
-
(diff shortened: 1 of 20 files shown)


duzos (approved):
LGTM

duzos:
however this isnt a "fix" its a "perf"

PR #1516, opened 2025-04-30
fix sonic modes making double calls

Mansarde:
## About the PR
Currently the sonic modes are being executed twice in a row.
In scanning mode, it prints the message with the Artron levels of the rift chunk **two times**.
In Overcharge or Interact mode it's a bit difficult to stop at the desired state of the targeted block.
For example if you try to turn a redstone lamp *on*, then if you stop the sonic right when the lamp turns on, the lamp gets toggled once more and turns *off*.

This PR fixes this, so that each sonic's mode-function is not toggled twice in a row anymore.

## Why / Balance
Better UX.
(I.e. no doubling up of messages or an additional toggling after sonic use has already stopped.)

## Technical details
The reason for the double-call is because both the `startUsing` and `tick` methods are called when the sonic is used.
First `startUsing` gets called and processes the respective mode-function.
Then `tick` gets called right afterwards and *also* processes the mode-function.

(Technically this also happens for the TARDIS mode, but it's not obvious because its messages are displayed on the action bar, where there is only ever one message visible. But it *is* being executed two times in a row.)

But for all the sonic modes, only `tick` is needed to process the sonic use, because even if the user clicks very shortly, `tick` is still called.
So we can leave out all `startUsing` (and `stopUsing`) calls, as they do nothing else than process the mode-function.
The result is that the mode-function is only executed once (per tick) and doesn't start (or stop) with a double-execution.

Which means that the Artron message in scanning mode will only display once.
And in overcharge or interaction mode you can now stop at the state you desire, without an extra execution throwing you off.

## Media
image
image
image
image

## Requirements

**Changelog**
:cl:
- fix: fix sonic modes makes double calls

### DIFF
--- a/src/main/java/dev/amble/ait/core/item/sonic/InteractionSonicMode.java
+++ b/src/main/java/dev/amble/ait/core/item/sonic/InteractionSonicMode.java
@@ -28,14 +28,6 @@ public class InteractionSonicMode extends SonicMode {
     }
 
     @Override
-    public void stopUsing(ItemStack stack, World world, LivingEntity user, int ticks, int ticksLeft) {
-        if (!(world instanceof ServerWorld serverWorld))
-            return;
-
-        this.process(serverWorld, user, ticks);
-    }
-
-    @Override
     public void tick(ItemStack stack, World world, LivingEntity user, int ticks, int ticksLeft) {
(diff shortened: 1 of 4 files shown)


duzos (approved):
LGTm

PR #1535, opened 2025-05-03
fix(README): fixed my nickname because nobody here can ever get it right

nitrogenez:
no bureaucracy.


Addi3 (changes requested):
LGTM 

edit: ignore this, im silly

Mansarde:
@nitrogenez Your name is also misspelled in https://github.com/amblelabs/ait/blob/main/CREDITS.md and https://github.com/amblelabs/ait/blob/main/src/main/resources/fabric.mod.json

Mansarde (approved):
LGTM

PR #1548, opened 2025-05-09
Various perf fixes

DrTheodor:
## About the PR
This is another performance PR with general performance fixes.

## Why / Balance
Because the performance SUCKS.

## Technical details
This PR makes it so the doors don't close on world save. I dont know why this was even a thing. Next, players are now restricted in block placement. They can only place blocks in the 500 block radius from the center of the world (x: 0, z: 0). 
Also this PR forces the server tardis manager to save the tardises only when the overworld gets saved (and not the other 100 miniworlds).

## Requirements

**Changelog**

:cl:
- perf: general performance improvements

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -82,7 +82,7 @@ dependencies {
         exclude(group: "net.fabricmc.fabric-api")
     }
 
-    modApi("me.shedaniel.cloth:cloth-config-fabric:${project.}") {cloth_version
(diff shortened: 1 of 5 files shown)


duzos (approved):
LGTM but i'm not sure on the 1000^2 limit

Addi3:
no, 500 it not happening chief

Addi3:
kill this man now, with hammers

PR #1551, opened 2025-05-11
Fix some controls being interact-able during growth state

### LINKED ISSUE #1525: [1.2.3] Protocol 515 usable in growth TARDIS
### What happened?

Protocol 515 can be used in a growth TARDIS with an engine + console to move it

### Version

1.2.1

### Build number & branch

_No response_

### Logs & screenshots

N/A

Mansarde:
## About the PR
This PR prevents interact-ability of the console controls during a TARDIS' growth-state.

fixes #1525 

## Why / Balance
The console controls should not be interactable during growth mode, as the TARDIS has not yet fully formed.

## Technical details
There is a guard-clause early on in the control-code, which returns false if the TARDIS is in the growth-state (so that neither the sound for a control plays, nor a missing-subsystem message is displayed, etc.).

## Requirements

**Changelog**
:cl:
- fix: controls were interact-able during growth state, but shouldn't

### DIFF
--- a/src/main/java/dev/amble/ait/core/blockentities/control/ControlBlockEntity.java
+++ b/src/main/java/dev/amble/ait/core/blockentities/control/ControlBlockEntity.java
@@ -100,9 +100,6 @@ public abstract class ControlBlockEntity extends InteriorLinkableBlockEntity {
         if (!(found.get() instanceof ServerTardis tardis))
             return false;
 
-        if (found.get().isGrowth())
-            return false;
-
         if (!this.control.canRun(tardis, user))
             return false;
 
--- a/src/main/java/dev/amble/ait/core/blockentities/control/ControlBlockEntity.java
+++ b/src/main/java/dev/amble/ait/core/blockentities/control/ControlBlockEntity.java
@@ -100,6 +100,9 @@ public abstract class ControlBlockEntity extends InteriorLinkableBlockEntity {
         if (!(found.get() instanceof ServerTardis tardis))
             return false;
 
+        if (found.get().isGrowth())
+            return false;
+
         if (!this.control.canRun(tardis, user))
             return false;
 
--- a/src/main/java/dev/amble/ait/core/entities/ConsoleControlEntity.java
+++ b/src/main/java/dev/amble/ait/core/entities/ConsoleControlEntity.java
@@ -378,14 +378,8 @@ public class ConsoleControlEntity extends LinkableDummyEntity {
     }
 
     public boolean run(PlayerEntity player, World world, boolean leftClick) {
-        Tardis tardis = this.tardis().get();
-
-        if (world.isClient() || tardis.isGrowth())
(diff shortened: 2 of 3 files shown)


src/main/java/dev/amble/ait/core/entities/ConsoleControlEntity.java
```diff
@@ -378,7 +378,9 @@ public void subtractDurability(float durability) {
     }
 
     public boolean run(PlayerEntity player, World world, boolean leftClick) {
-        if (world.isClient())
+        Tardis tardis = this.tardis().get();
+
+        if (world.isClient() || tardis.isGrowth())
```
Mansarde, line 383: I whould probably put the growth check in its own condition.
But I wasn't sure how you would want it, so I opted for keeping it compact.

Let me know if you want this to be separate. :)

DrTheodor: nitpick: you should first do a `isClient` check and only then get the tardis.
the cost of getting the tardis is more than checking if the world is on client

src/main/java/dev/amble/ait/core/blockentities/control/ControlBlockEntity.java
```diff
@@ -100,6 +100,9 @@ public boolean run(ServerPlayerEntity user, boolean isMine) {
         if (!(found.get() instanceof ServerTardis tardis))
             return false;
 
+        if (found.get().isGrowth())
```
DrTheodor, line 103: i feel like this could be moved inside `Control#canRun`

src/main/java/dev/amble/ait/core/entities/ConsoleControlEntity.java
DrTheodor: also check my previous comment on ``ControlBlockEntity`

DrTheodor (changes requested):

src/main/java/dev/amble/ait/core/blockentities/control/ControlBlockEntity.java
Mansarde: Ok, I can do that.
The reason I put it into `run` for both `ControlBlockEntity` and `ConsoleControlEntity` was because in the latter I wanted it to prevent the secret music from playing if the TARDIS is still in growth:
https://github.com/amblelabs/ait/blob/99da8d9b0773293e7b313469260539c21b95fa05/src/main/java/dev/amble/ait/core/entities/ConsoleControlEntity.java#L386-L388

But I could move the secret music down, right after `canRun` instead I suppose.
Will do that and test it. :)

Mansarde:
@DrTheodor Ok, I've moved the check to the parent class and restored the changes in the previous two console classes, which should resolve both of your requested changes.

DrTheodor (approved):
LGTM

PR #1503, opened 2025-04-28
Sticky keys

MaketendoDev:
## About the PR
added the option to prank your friends by right clicking a slimeball on the control to lock it and shears to unlock it.

## Why / Balance
it makes it more interactive

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://amblelabs.github.io/ait-wiki/guidelines).
- [X] I have added media to this PR or it does not require an ingame showcase.

**Changelog**

PR #1553, opened 2025-05-12
fix sticky controls being bypass-able via left-click

Mansarde:
## About the PR
This PR fixes the sticky-controls feature (#1503) being bypassed when punching (i.e. left-clicking) controls.

## Why / Balance
When controls are being made sticky, then it shouldn't matter if we left- or right-click them.
They should prevent the control from operating in either case.

## Technical details
(The mentioned methods all relate to [ConsoleControlEntity](https://github.com/amblelabs/ait/blob/main/src/main/java/dev/amble/ait/core/entities/ConsoleControlEntity.java))

When interacting (i.e. right-clicking) with a control, the `interact` method is called, at the end of which the `run` method is called (which executes the main functionality of the control).
The `interact` method handles sticky controls and prevents `run` from being called.
But when punching (i.e. left-clicking) a sticky control, this calls `run` directly and thus bypasses `interact`, which means the stickiness will be bypassed as well.

This PR moves the stickiness logic from the `interact` method into the `run`method.
I also set `true` for its various returns, so that their own sound plays.
Otherwise when `run` returns from its call in `interact`, the fail sound-FX would play and mask the stickiness-sound FX.

## Requirements

**Changelog**

:cl:
- fix: sticky keys could be bypassed via punching / left-clicking controls

### DIFF
--- a/src/main/java/dev/amble/ait/core/entities/ConsoleControlEntity.java
+++ b/src/main/java/dev/amble/ait/core/entities/ConsoleControlEntity.java
@@ -192,27 +192,6 @@ public class ConsoleControlEntity extends LinkableDummyEntity {
             return ActionResult.SUCCESS;
         }
 
-        if (isSticky()) {
-            if (player.getMainHandStack().isOf(Items.SHEARS)) {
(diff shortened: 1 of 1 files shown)


duzos (approved):
LGTM

PR #1563, opened 2025-05-26
fix: handle world load failure better

DrTheodor:
## About the PR
This PR improves the multidim world load failure by creating a new world instead of silently failing.

Also, it removes the multidim persistence.

## Why / Balance
If multidim fails to save the (useless) world data, it will fail to load it afterwards, even though AIT can provide the same blueprint data with no troubles whatsoever.

## Requirements

**Changelog**

🆑
- fix: make multidim less error-prone


duzos (approved):
LGTM

PR #1564, opened 2025-05-26
fix: no more clientside subsystem creation

drtheodor:
## About the PR
This PR changes the subsystem handler to not create subsystems on get.

## Why / Balance
Previously, this could become a reason for client-side crashes: the engine would get the subsystems to load, the handler would get the subsystem and see that its null, then it would try initializing one and putting it in the map and syncing it. to the server.

## Technical details
This PR removes the ability to remove subsystems, as well as reworking the subsystem iterator.

## Requirements

## Breaking changes
- `SubSystemHandler#remove` removed. Use `SubSystem#setEnabled` instead.
- All subsystems provided by the handler are now non-null by default.

**Changelog**

🆑
- fix: no more crashes related to subsystems trying to sync themselves to the server


duzos (approved):
LGTM

PR #1565, opened 2025-05-26
fix: siege held uuid backwards compat

DrTheodor:
## About the PR
This PR fixes the backwards compatibility between the new and one of the older releases.

## Why / Balance
One of the previous PRs allowed to use `null` values in properties. This PR fixes the backwards compatibility of the TARDIS' created before that release.

## Technical details
`SiegeHandler` was changed to use the null value to check whether the sieged tardis is being held or not, but the older versions used `new UUID(0, 0)`.

## Requirements

**Changelog**

🆑
- fix: siege should be held-able with TARDIS' created prior to one of the previous releases


duzos (approved):
LGTM

PR #1572, opened 2025-05-29
Durability for Stat Remotes

Loqor:
## About the PR
Added durability to stat remotes.

## Why / Balance
They're extremely unbalanced.

## Technical details
Added 300 durability to them.

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/src/main/java/dev/amble/ait/core/AITItems.java
+++ b/src/main/java/dev/amble/ait/core/AITItems.java
@@ -59,7 +59,7 @@ public class AITItems extends ItemContainer {
     // Functional Items
     @NoEnglish
     public static final Item REMOTE_ITEM = new RemoteItem(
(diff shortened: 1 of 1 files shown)


smallmoss233:
Stupid idea, the Stenheim remote control is so difficult to obtain, I think it's worth it if it doesn't have durability.

Addi3:
ehh, it is the most OP item in the mod and its not massively expensive

smallmoss233:
The high manufacturing cost should be worth a sufficiently high return. The Stenheim Remote is already marginalized in survival, and now this move has nailed it to the tracks of death.

smallmoss233:
Compared to summoning the slow TARDIS, I choose to set my respawn point inside the TARDIS. The Stenheim Remote? Sorry, it's not worth it.

PR #1578, opened 2025-06-05
feat: scan exterior in overload to cancel demat / toggle alarms etc

duzos:
## About the PR

If you now scan a tardis you are loyal to in overload mode, certain controls will run.
For example, a TARDIS which is dematting will flick the handbrake to cancel it.

## Why / Balance

Funsies

## Requirements

### DIFF
--- a/src/main/java/dev/amble/ait/core/item/sonic/OverloadSonicMode.java
+++ b/src/main/java/dev/amble/ait/core/item/sonic/OverloadSonicMode.java
@@ -1,14 +1,26 @@
 package dev.amble.ait.core.item.sonic;
 
+import dev.amble.ait.api.tardis.link.v2.Linkable;
+import dev.amble.ait.api.tardis.link.v2.block.AbstractLinkableBlockEntity;
 import dev.amble.ait.core.AITSounds;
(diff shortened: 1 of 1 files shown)


src/main/java/dev/amble/ait/core/item/sonic/OverloadSonicMode.java
```diff
@@ -57,6 +69,16 @@ private void overloadBlock(BlockPos pos, ServerWorld world, LivingEntity user, i
         if (!canMakeRedstoneTweak(ticks))
             return;
 
+        if (world.getBlockEntity(pos) instanceof AbstractLinkableBlockEntity ext && ext.isLinked()) {
```
DrTheodor, line 72: for performance reasons, it'd be better to check whether the `block` is actually an `ExteriorBlock` before doing `World#getBlockEntity`

DrTheodor (changes requested):

src/main/java/dev/amble/ait/core/item/sonic/OverloadSonicMode.java
duzos: thats what i wanted

PR #1621, opened 2025-06-15
pt_br translation update

kluas005:
## About the PR

I've been reviewing it, it seems like there's always something that doesn't look so good and needs to be changed, but anyway, I think it's better for now, I'll probably only change something when they add more things
## Requirements

- [ ] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

- added/changed Portuguese translations


Mansarde (approved):
LGTM

PR #1635, opened 2025-06-19
chore: add BETA to version dropdown

Mansarde:
## About the PR
Added ~1.2.8~ `BETA` to the bug report version list

### DIFF
--- a/.github/ISSUE_TEMPLATE/1-bug-report.yaml
+++ b/.github/ISSUE_TEMPLATE/1-bug-report.yaml
@@ -29,6 +29,7 @@ body:
     label: Version
     description: What version of the mod you're running?
     options:
+    - BETA
     - 1.1.0
     - 1.2.0
     - 1.2.1
@@ -38,7 +39,6 @@ body:
     - 1.2.5
     - 1.2.6
     - 1.2.7
-    - 1.2.8
     default: 0
   validations:
     required: true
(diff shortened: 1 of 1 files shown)


Mansarde:
In case nobody is approving because there is no 1.2.8 release yet:
We also have to keep in mind that betatesters test with versions that can be higher than release versions (as is the case right now).

MaketendoDev:
I think it would be better to instead just mark it down as BETA than 1.2.8 as you don't know just yet if we need to do a emergency patch or somthing.

MaketendoDev:
I think it would be better to instead just mark it down as BETA than 1.2.8 as you don't know just yet if we need to do a emergency patch or something.

Mansarde:
So according to the bug template so far, you first choose the release version from the dropdown, but you can then specify a build number if it's *not* a release version.
If you left the build version off, then it's assumed to the a release version.

image

But I can change the PR so that we always just have "BETA" in the dropdown, is that what you mean?

DrTheodor:
Yes please 
The thing about betas is that there can be different branches of them
So different fixes and so on 
So if a bugreport for a beta build appears without a build number it can be as good as useless

Mansarde:
Ok @DrTheodor @MaketendoDev, I changed it to add `BETA` instead.

DrTheodor (approved):

Issue #1637, opened 2025-06-19
Null crash related to sonic scanning


Saturnorsomthing:
### What happened?

Scanning a bell/zeiton with a sonic that does not have a linked TARDIS results in a null crash.

i was gonna fix it myself but i realized i do not have time so uhh whoopsie
just add a null check

### Version

1.2.7

### Logs & screenshots

e

### I've checked the reported issues...

- [x] ...and this issue was not reported before.

Mansarde:
K, I'll take a look.

Mansarde:
Resolved with #1639

Issue #1636, opened 2025-06-19
Sonic Scanning giving TARDIS Dimension is wrong


Saturnorsomthing:
### Description

Using the sonic in scanning mode on zeiton/bells shows the correct cordinates, however the dimension is not the one the TARDIS is in but rather the player

### I've checked the feature requests...

- [x] ...and this feature request was not requested before.

Mansarde:
Huh, I thought that was fixed already.
May I ask on which version?

Saturnorsomthing:
latest im fairly sure
i might be schizophrenic give me a second

Mansarde:
Resolved with #1639

PR #1639, opened 2025-06-19
Fix sonic locator bugs

Mansarde:
## About the PR
Fixes two bugs with the sonic's TARDIS locator functionality, as reported in #1636 and #1637:

- The message incorrectly shows the *player's* current dimension instead the dimension the TARDIS is located in.
- And the game crashes due to a NPE when the TARDIS linked to the sonic doesn't exist anymore.

## Why / Balance
Because less bugs > more bugs

## Technical details
It's two very small changes that should be apparent from the diff itself.
Basically I just included a null check for the TARDIS object, as well as replaced the use of `world` with the one derived from the TARDIS object.

## Requirements

**Changelog**
:cl:
- fix: Sonic's locator function crashed the game when the linked TARDIS didn't exist anymore.
- fix: Sonic's locator incorrectly showed the player's dimension instead the TARDIS' one.


Loqor (approved):
LGTM

PR #1646, opened 2025-06-21
Fix ability to circumvent valid Y-coordinate ranges for dimensions

Mansarde:
## About the PR
This PR fixes the ability to make the TARDIS land beyond the valid maximum Y coordinate of a dimension, as well as beyond the worldborder coordinates.

<details>
<summary>Details</summary>

When entering a Y coordinate via the console controls, its range of valid coordinates will be determined by the current dimension target. But after switching to a new dimension target, the Y is not automatically adjusted to the valid range for the newly chosen dimension.
(Only after clicking the Y control again will the coordinates be adjusted.)

This can lead to e.g. being able to travel above the valid heights of dimensions (e.g. on top of the Nether roof, above the build height of the Overworld, etc.)

For example one could set the dimension to "Space", enter the max Y of 2031 and then switch to the Overworld dimension.
As long as the Y control isn't touched again, the target Y coordinate will stay at 2031 and the TARDIS *will* land there (if the vertical land type isn't enabled).
This would lead the TARDIS exterior to not be rendered and also not fall to the ground.
</details>

## Why / Balance
Because one is not supposed to circumvent these limitations.
Otherwise people TARDISes could end up above the Nether roof, above the build height or outside the world border.

## Technical details
The Y control has a block of code that clamps the Y coordinate depending on the target dimension.
I extracted that block of code into its own static function and use that for both the Y control, and now for the dimension control as well.
Additionally I make the dimension control now use `TravelHandler::destination` instead of `TravelHandler::forceDestination`, to make it honor the worldborder too.

## Requirements

**Changelog**
:cl:
- fix: The dimension control can no longer be used to land outside the valid Y range or outside the worldborder.

### DIFF
--- a/src/main/java/dev/amble/ait/core/tardis/control/impl/DimensionControl.java
+++ b/src/main/java/dev/amble/ait/core/tardis/control/impl/DimensionControl.java
@@ -3,6 +3,7 @@ package dev.amble.ait.core.tardis.control.impl;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 
+import dev.amble.ait.core.tardis.control.impl.pos.PosType;
 import dev.amble.lib.data.CachedDirectedGlobalPos;
 
 import net.minecraft.server.network.ServerPlayerEntity;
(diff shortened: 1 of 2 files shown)


Loqor (approved):
LGTM

PR #1691, opened 2025-07-02
fix the so-called "impossible" task of... making scrolling text.

Loqor:
## About the PR
Added scrolling text to the monitors to address stuff like names being too long or for some reason dimension names.

## Why / Balance
Fixing a common player grievance.

## Technical details
Just added a better version of our already-existing truncating method in MonitorUtil.

## Media
https://cdn.discordapp.com/attachments/1213995158178242631/1389852922157858986/2025-07-02_01-18-32.mp4?ex=686620e6&is=6864cf66&hm=624d5c3a2b269f572151b13da3516fa0ed13386bd20ca028794342e37210c838&

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: scrolling text on monitors for long strings of text.

### DIFF
--- a/src/main/java/dev/amble/ait/client/renderers/monitors/MonitorRenderer.java
+++ b/src/main/java/dev/amble/ait/client/renderers/monitors/MonitorRenderer.java
@@ -101,52 +101,66 @@ public class MonitorRenderer<T extends MonitorBlockEntity> implements BlockEntit
         BlockPos abpdPos = abpd.getPos();
 
         String positionPosText = " " + abppPos.getX() + ", " + abppPos.getY() + ", " + abppPos.getZ();
-        Text positionDimensionText = Text.of(MonitorUtil.truncateDimensionName(WorldUtil.worldText(abpp.getDimension()).getString(), 20));
-
-        this.textRenderer.drawWithOutline(Text.of("❌").asOrderedText(), 0, 0, 0xF00F00, 0x000000,
-                matrices.peek().getPositionMatrix(), vertexConsumers, 0xF000F0);
-        this.textRenderer.drawWithOutline(Text.of(positionPosText).asOrderedText(), 0, 8, 0xFFFFFF, 0x000000,
-                matrices.peek().getPositionMatrix(), vertexConsumers, 0xF000F0);
-        this.textRenderer.drawWithOutline(positionDimensionText.asOrderedText(), 0, 16, 0xFFFFFF, 0x000000,
-                matrices.peek().getPositionMatrix(), vertexConsumers, 0xF000F0);
-        this.textRenderer.drawWithOutline(WorldUtil.rot2Text(abpp.getRotation()).asOrderedText(), 0, 24, 0xFFFFFF, 0x000000,
-                matrices.peek().getPositionMatrix(), vertexConsumers, 0xF000F0);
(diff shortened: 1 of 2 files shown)


duzos:
loop is prolly fine

duzos (changes requested):
@Mansarde comment

Mansarde:
Actually, ignore my last comment, I've figured out a way to do it without the loop.
I'll add it on top of the scrollText PR once it's done.

src/main/java/dev/amble/ait/core/util/MonitorUtil.java
```diff
@@ -1,10 +1,50 @@
...
     public static String truncateDimensionName(String name, int maxLength) {
         if (name.length() > maxLength) {
             return name.substring(0, maxLength) + "...";
         }
         return name;
     }
+
+    // Use this one instead, it scrolls! - Loqor
+    public static String scrollText(String name, int maxLength, int scrollIndex, float tickDelta) {
+        time += tickDelta;
```
Mansarde, line 16: Since `time` is never reset, wouldn't it grow unimpeded?

src/main/java/dev/amble/ait/client/renderers/monitors/MonitorRenderer.java
```diff
@@ -101,52 +101,66 @@ public void render(MonitorBlockEntity entity, float tickDelta, MatrixStack matri
...
 
         String name = tardis.stats().getName();
-        this.textRenderer.drawWithOutline(Text.of(name).asOrderedText(), 98 - (this.textRenderer.getWidth(name)), 90,
-                0xFFFFFF, 0x000000, matrices.peek().getPositionMatrix(), vertexConsumers, 0xF000F0);
 
-        if (tardis.alarm().isEnabled())
+        int lightLevel = 0xF000F0;
+
+        // Position
+        drawTextWithIcon(matrices, vertexConsumers, "❌", positionPosText, 0, 0, 0xF00F00, 0xFFFFFF, overlay, lightLevel);
```
Mansarde, line 120: Drawing text with an icon next to it reduces the code a bit which is nice, but doesn't this also mean the coordinates now have less space?

Before, the coordinates where below the icon, like so:
```
❌
-12345, 250, -32105
```

But now they would be on the same line, leaving less room for the coordinates to fit on the monitor screen:
```
❌ -12345, 250, -32105
```

@Loqor 
Since I'm working on the same class at the moment, would it be ok if I modified the `drawTextWithIcon` in my branch to separate the icon and text into separate lines?
This way there'd still be the upside of less code (by using `drawTextWithIcon`), but without losing space for the coordinates.

Mansarde: Or actually, since I see it's used for the fuel and travel progress as well and makes sense to be in one line there, I'll instead just overload `drawTextWithIcon` to allow for manually setting the X and Y offsets, as an option.

Ignore my previous poste then, ^^

Mansarde:
Wtf, this was *not* ready for merge!
See my comment here with the problems: https://github.com/amblelabs/ait/pull/1691#issuecomment-3026713501
It was only ever tested on the default TARDIS name and doesn't work for any other length of name.

Mansarde:
Wait, it got merged but *didn't* end up in main?
I'm confused. Gotta check Discord to see what's happened...

PR #1724, opened 2025-07-19
updated exile variant texture

TheMilkSnatcher:
it was old and needed to be redone lool

mainly fixed the PCB signs and a line of miscoloured pixels on one of the walls

(let SWD also get a hand of this since it also uses this texture for the AiT box in DM)

**Changelog**
:cl:
- change: updated exile variant texture


Mansarde (approved):
LGTM

PR #1726, opened 2025-07-19
fix warning of interior change cost

Mansarde:
## About the PR
This PR includes the instability in the fuel-check for interior changes.

## Why / Balance
The interior change reduces the fuel by 5000 Artron multiplied by the instability.
But the check that displays a warning when there is too little fuel only checks against 5000 Artron.
That means that as long as someone has >= 5000 Artron, the interior change will be successful.
But this is exactly what the fuel check is supposed to prevent, i.e. one should need 5000 x instability of Artron.

## Technical details
This PR simply adds the TARDIS' current instability value as a factor to the fuel cost (i.e. `5000 * tardis.travel().instability()`.
It also converts the base Artron cost to a constant (i.e. `MIN_FUEL_COST` instead of `5000`).

## Requirements

**Changelog**
:cl:
- fix: the check for interior change did not include the increased fuel cost due to instability


DrTheodor (approved):
LGTM

PR #1738, opened 2025-07-23
fix lingering waterlogged property of door

Mansarde:
## About the PR
This PR fixes the door remaining waterlogged after closing it or activating the shields.

## Why / Balance
Because there should be no water continuing to flow inside the TARDIS after the doors have been closed or the shields activated.

## Technical details
I'm not 100% confident about the locations where I disable the waterlogged property.
But I tried to place it as close to the final "close door" and "activate shields" places in the codebase as I could find.

I originally added it to the tick method of the `DoorBlockEntity`, but I thought checking this every tick was unnecessary and it would suffice to remove the waterlogged property only whenever the door-close or shield-activate events occurred.

## Requirements

**Changelog**
:cl:
- fix: water (from landing underwater and opening the doors) kept flowing into the TARDIS despite closing the doors or activating the shields

### DIFF
--- a/src/main/java/dev/amble/ait/core/tardis/handler/DoorHandler.java
+++ b/src/main/java/dev/amble/ait/core/tardis/handler/DoorHandler.java
@@ -1,7 +1,11 @@
 package dev.amble.ait.core.tardis.handler;
 
+import dev.amble.ait.core.tardis.Tardis;
 import dev.amble.lib.data.DirectedBlockPos;
 import net.fabricmc.fabric.api.util.TriState;
+import net.minecraft.block.Block;
+import net.minecraft.block.BlockState;
+import net.minecraft.state.property.Properties;
 import org.jetbrains.annotations.Nullable;
 
 import net.minecraft.particle.ParticleEffect;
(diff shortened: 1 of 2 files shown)


Mansarde:
> not the best solution, but alright

Yeah I wasn't too confident about it. I kept in mind not to add it to the tick method and keep it as close to the door-close and shield events as I was able to.
I'm always open for suggestions though, can you give a hint on how it could be improved?

DrTheodor:
Well, the thing that concerns me the most is that you split the (block) state handling, so now it's shared between the door block entity code and door handler. Ideally, it should only be in one of those. That could be done by either moving all the waterlogged door logic to DoorHandler or moving it all to the block entity. To avoid checking updates every tick you could use events, and since the door positions are known from a TARDIS instance, it could be done that way.

In the ideal world, DoorHandler would manage all the waterlogged state and send an event that "a door got waterlogged!", in which case the loyalty handler would manage the logic for the loyalty depletion on itself. That would remove a hard dependency on loyalty handler, remove TARDIS logic (besides linking, teleporting and stuff) from the door, and move all the door related state handling to the DoorHandler.

Mansarde:
Hmm, yeah that would require some rewriting, but I see your point.
The way it is currently is really just a quick direct fix to prevent the waterlogged annoyance that people had to deal with.
But I'll keep what you suggested in mind and try to get it to work that way, so it's less coupled for the future.

Thanks for the input, much appreciated! 🍻

PR #1739, opened 2025-07-23
chore: re-ballanced loyalty stuff

Saturnorsomthing:
## About the PR

Adjusted some values related to loyalty draining

## Why / Balance

The loyalty drainage was never changed even after loyalty was changed from 100 to 500, meaning some things were unfair or unballanced as you lose loyalty fast but gain it very slowly.

## Technical details

Adjusted a few values here and there and copy n pasted a line upwards.

## Requirements

## Breaking changes

**Changelog**

:cl:
- Re-balanced loss of loyalty.


Mansarde (approved):
So, less harsh when letting in water or destroying keys, but more harsh when hammering the TARDIS. Makes sense, since the latter is quite intentional harm directed at the TARDIS.

I'd be more mad as well if you hit me with a hammer and less so if you just lost my keys or sprayed water onto me. 😆

LGTM

DrTheodor:
booooooooooo i dont agree

PR #1757, opened 2025-07-26
fix: extreme shaking

### LINKED ISSUE #1755: spazzing out screen when caching console during flight
### What happened?

This happened!

1. make a tardis and select the hudolin or another console
2. proceed to take off
3. when during flight cache your console
4. screen starts to spazz out like crazy until you select a console again
this is on the latest hudolin release

### Version

1.2.10

### Mod Loader

fabric

### Build number & branch

ait-1.2.11.2825-1.20.1-release

### Logs & screenshots

_No response_

DrTheodor:
## About the PR
This PR fixes extreme shaking as seen in #1755. Fixes #1755.

## Why / Balance
Because such screen shaking makes the game unplayable.

## Technical details
Shaking intensity during flight depends on the distance from the console. If there's no console then the method `getDistanceToConsole` will return `Double.MAX_VALUE`, which somewhat makes sense logically, but breaks the logic. This PR updates this method to return 0 in case there are no consoles.

Although returning 0 doesn't make sense logically, this method returns 0 if it encounters an error (like, if the client player is not in a tardis, or, if theres, for some reason, the lack of the tardis object).

## Requirements

## Breaking changes
- `ClientTardisUtil#distanceFromConsole` does no longer return `Double.MAX_VALUE` if there are no consoles.

**Changelog**

:cl:
- fix: no more extreme screen shaking when there are no consoles during flight

### DIFF
--- a/src/main/java/dev/amble/ait/client/util/ClientTardisUtil.java
+++ b/src/main/java/dev/amble/ait/client/util/ClientTardisUtil.java
@@ -182,10 +182,15 @@ public class ClientTardisUtil {
         if (tardis == null)
             return 0;
 
+        Collection<BlockPos> consoles = tardis.getDesktop().getConsolePos();
+
+        if (consoles.isEmpty())
+            return 0;
+
         BlockPos pos = player.getBlockPos();
         double lowest = Double.MAX_VALUE;
-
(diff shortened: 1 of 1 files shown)


Mansarde (approved):
Successfully tested and LGTM. 🙂👍

Issue #1701, opened 2025-07-08
waypoint deletes itself when a new waypoint is put in its place [instead of just dropping the waypoint cartridge or loading a "this slot is occupied" message)


Dogesarecute:
### What happened?

waypoint overwrites itself when clicking with a new waypoint cartridge
1. Get a waypoint cartridge
2. Click on a waypoint bank [that already has lets say 4 cartridges installed]
3. Place the waypoint in a slot close to a pre existing one
4. if you missclick it deletes your waypoint and puts the waypoint you placed in that same slot [if that makes sense]

### Version

1.2.7

### Mod Loader

forge

### I've checked the reported issues...

- [x] ...and this issue was not reported before.

Mansarde:
I was able to replicate this, both in the reported mod version 1.2.7 and the latest release 1.2.11
That is, trying to insert a new cartridge into the slot already occupied by another cartridge will replace the old cartridge with the new one.

PR #1762, opened 2025-07-28
Fix/port eating cartridges and throwing exceptions

Mansarde:
## About the PR
This PR fixes two issues:

1) Exceptions:
When the player tries to eject an item from either the sonic- or the console-port when the respective port is *empty*, then an exception is thrown (though the game doesn't crash).

2) Console eating waypoint cartridges:
When trying to eject a waypoint cartridge after swapping in another cartridge (i.e. without first ejecting the first one), then the previous cartridge will not actually be ejected and instead be gone for good.

## Why / Balance
The exceptions don't crash the game, but should be accounted for with null checks.
And the console destroying waypoint cartridges (after swapping in another cartridge) is unexpected behavior and could lead to unintended loss of cartridges.

## Technical details
For the exceptions, simple null checks have been added.
To prevent the console from eating/destroying cartridges, the `clearCartridge()` call has been moved to the `clear` method, so that the `hasCartridge` state doesn't become false in the case when a player is swapping in a different cartridge.

## Requirements

**Changelog**
:cl:
- fix sonic- and console-port throwing exceptions
- fix console part eating cartridges when swapping

### DIFF
--- a/src/main/java/dev/amble/ait/core/tardis/control/impl/SonicPortControl.java
+++ b/src/main/java/dev/amble/ait/core/tardis/control/impl/SonicPortControl.java
@@ -43,6 +43,9 @@ public class SonicPortControl extends Control {
                 item = butler.takeHandles();
             }
 
+            if (item == null)
+                return Result.FAILURE;
+
             player.getInventory().offerOrDrop(item);
             consoleBlockEntity.setSonicScrewdriver(ItemStack.EMPTY);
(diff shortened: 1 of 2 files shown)


rapbattlegod32:
https://github.com/amblelabs/ait/issues/1701 does this pr link to this issue?

rapbattlegod32 (approved):
LGTM everythings working as it should

PR #1764, opened 2025-07-28
Fix cartridges replacing occupied slots in waypoint banks

Mansarde:
## About the PR
This PR prevents cartridges in waypoint banks from being overridden/destroyed when inserting another cartridge in an already occupied slot.

## Why / Balance
Neither the waypoint bank, not cartridges themselves should be able to eat cartridges. :P

## Technical details
A check was added to the `insert` method of the waypoint bank blockentity to prevent the insertion if the slot is not null.

## Requirements

**Changelog**
:cl:
- fix occupied slots in waypoint banks from being overridden when trying to insert another cartridge there


rapbattlegod32 (approved):
LGTM

Issue #1613, opened 2025-06-14
Landing Marker spots floating in air


Ouroboros-Games:
### What happened?

The parking spot indicators the landing marker shows when you have a sonic in scanning mode in your hand are always floating 1 block above their actual position

### Version

1.2.6

### Logs & screenshots

N/A

### I've checked the reported issues...

- [x] ...and this issue was not reported before.

PR #1767, opened 2025-07-29
fix: landing pad issues

### LINKED ISSUE #1612: Landing Marker spots can no longer be changed (1.2.0+)
### What happened?

Whereas previously a sonic in scanning mode could create/remove the parking spots on a Landing Marker, it can no longer do that

### Version

1.2.6

### Build number & branch

_No response_

### Logs & screenshots

N/A

duzos:
closes #1612 and #1613

🆑 
fix: unable to move landing pad spots
fix: spots rendering at the wrong position
fix: landing marker failing in unloaded chunks

### DIFF
--- a/src/main/java/dev/amble/ait/core/world/LandingPadManager.java
+++ b/src/main/java/dev/amble/ait/core/world/LandingPadManager.java
@@ -43,7 +43,7 @@ public class LandingPadManager {
     }
 
     @Nullable public LandingPadRegion getRegion(ChunkPos pos) {
-        Chunk chunk = this.world.getChunk(pos.x, pos.z, ChunkStatus.FULL, false);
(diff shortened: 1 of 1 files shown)


Mansarde (approved):
LGTM

Tested the following:
- Try to summon TARDIS into landing pad chunk without correct password.
=> Result: Fail (as expected)
- Try to summon TARDIS into landing pad chunk with correct password.
=> Result: Success (as expected)
- Try to fly into landing pad chunk without correct password.
=> Result: Fail (as expected)
- Try to fly into landing pad chunk with correct password.
=> Result: Success (as expected)
- All the same as above, but with removing all spots except one
=> Result: Success, it landed on the only available spot (as expected)
- Reloading map and checking if password is still set and working.
=> Result: Success (as expected)

tl;dr
So far everything seems to work as expected. :)

PR #1769, opened 2025-07-29
fix: the load waypoint control should load the coordinates, not start travel

Mansarde:
## About the PR
Makes the "load waypoint" control only load the waypoint instead of instantly traveling.

## Why / Balance
This brings it in line with how the waypoint bank works, where confirming a pulled cartridge would simply load the coordinates.
It's also more intuitive, as someone would probably expect the console port and load button to work like a floppy drive or similar.
And it mirrors the save function, i.e.:
Put in cartridge -> press save -> coordinates saved
Put in cartridge -> press load -> coordinates loaded

## Technical details
First I renamed all the old occurrences of "mark" and "set" to "save" and "load" to avoid future confusion.
Then for the actual functional change I combined the methods `gotoWaypoint()` and `setDestination()` into the new method `loadWaypoint()`, which does mostly the same as `gotoWaypoint()` did, except that it doesn't initiate travel, but instead just sets the destination.

## Requirements

**Changelog**
:cl:
- fix: loading of waypoint cartridges now only loads the coordinates instead of starting flight


Mansarde:
Done, ready for review.

duzos (approved):
LGTM

PR #1770, opened 2025-07-29
sonic screwdriver summon notify if no power

rapbattlegod32:
## About the PR
if the tardis has no power and you try and summon it, it will now send a message to the player saying the power is off

## Why / Balance
image

## Requirements

**Changelog**

:cl:
- add sonic screwdriver summon notify if no power


Mansarde (approved):
LGTM

But just as a suggestion, if you agree:
"doesn't have power" could be misinterpreted as the TARDIS missing fuel / Artron.
Perhaps being more explicit by saying "is powered off" might be less ambiguous.

DrTheodor (approved):

PR #1775, opened 2025-07-30
Added the Axe thing Theo told me also made it mod compatible

Lt-DO7:
## About the PR

The axe breaking powered down locked TARDIS code.
## Why / Balance

It will make wooden and stone axes not able to break down the doors making it more of a challenge and Theo told me to impliment it.
## Technical details

 if (stack.getItem() instanceof AxeItem axeItem && axeItem.getAttackDamage()>=8 && stack.getItem() != Items.STONE_AXE) {
## Requirements

- [ X] I have read and am following the [Pull Request and Changelog Guidelines](https://amblelabs.github.io/ait-wiki/guidelines).
- [X ] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

<!--
:cl:
- remove: wooden and stone axes being able to break open TARDIS doors.
- add: iron and upwards being able to break TARDIS doors open + mod compatability too.

### DIFF
--- a/src/main/java/dev/amble/ait/core/tardis/handler/FuelHandler.java
+++ b/src/main/java/dev/amble/ait/core/tardis/handler/FuelHandler.java
@@ -5,6 +5,7 @@ import dev.amble.lib.data.CachedDirectedGlobalPos;
 import net.minecraft.block.BlockState;
 import net.minecraft.item.AxeItem;
 import net.minecraft.item.ItemStack;
+import net.minecraft.item.Items;
 import net.minecraft.server.MinecraftServer;
 import net.minecraft.server.world.ServerWorld;
 import net.minecraft.sound.SoundCategory;
@@ -64,7 +65,7 @@ public class FuelHandler extends KeyedTardisComponent implements ArtronHolder, T
             }
 
             // if holding an axe then break open the door RAHHH
-            if (stack.getItem() instanceof AxeItem) {
+            if (stack.getItem() instanceof AxeItem axeItem && axeItem.getAttackDamage()>=8 && stack.getItem() != Items.STONE_AXE) {
                 if (tardis.siege().isActive())
                     return DoorHandler.InteractionResult.CANCEL;
 


src/main/java/dev/amble/ait/core/tardis/handler/FuelHandler.java
```diff
@@ -64,7 +65,7 @@ public class FuelHandler extends KeyedTardisComponent implements ArtronHolder, T
             }
 
             // if holding an axe then break open the door RAHHH
-            if (stack.getItem() instanceof AxeItem) {
+            if (stack.getItem() instanceof AxeItem axeItem && axeItem.getAttackDamage()>=8 && stack.getItem() != Items.STONE_AXE) {
```
Mansarde, line 68: If the attack damage of the axe is >= 8, then the current order of checks will always evaluate both the attack damage *and* whether it's a stone axe.

I would swap these two checks though, to make the condition fail earlier.
That way, if we *have* a stone axe, the condition can already fail and not evaluate the attack damage, as that is redundant.
Only if we don't have a stone axe will the attack damage need to be evaluated.

Mansarde (changes requested):

Mansarde (approved):
LGTM

src/main/java/dev/amble/ait/core/tardis/handler/FuelHandler.java
BredDino: Just want to clarify for some reason the given attack damage is 1 lower than the actual attack damage so even tho the iron axe has 9 attack damage it only worked when he set it to 8

Mansarde: Hmm, perhaps the attack damage starts at 0 internally, but is *displayed* to the player incremented?
Haven't looked at the vanilla MC code, and I would usually expect this only when displaying index values in-game, but not with a variable that directly holds the value.

Ah well, good to know either way, thx for mentioning! 👍

PR #1780, opened 2025-07-31
Allow the use of default cave interior without breaking growth checks

DrTheodor:
## About the PR
This PR makes the already available growth desktop to be used without borking your TARDIS.

## Why / Balance
It's a pretty nice interior and people have tried to use it only to bork their TARDIS.

## Technical details
"Is TARDIS in growth?" check checks whether the exterior is a TARDIS coral OR the interior is the cave interior. This PR removes the second condition.

## Requirements

**Changelog**

:cl:
- fix: TARDIS' no longer break when using the default cave interior

### DIFF
--- a/src/main/java/dev/amble/ait/core/tardis/Tardis.java
+++ b/src/main/java/dev/amble/ait/core/tardis/Tardis.java
@@ -182,7 +182,7 @@ public abstract class Tardis extends Initializable<TardisComponent.InitContext> 
 
     // FIXME: this needs to be changed.
     public boolean isGrowth() {
-        return hasGrowthExterior() || hasGrowthDesktop();
+        return hasGrowthExterior();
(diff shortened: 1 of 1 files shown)


src/main/java/dev/amble/ait/core/tardis/Tardis.java
```diff
@@ -182,7 +182,7 @@ public boolean isUnlocked(Unlockable unlockable) {
 
     // FIXME: this needs to be changed.
     public boolean isGrowth() {
-        return hasGrowthExterior() || hasGrowthDesktop();
+        return hasGrowthExterior();
     }
```
Mansarde, line 186: Is the change you made here the FIXME that needed changing?
If so we probably don't need that reminder anymore.

PR #1789, opened 2025-08-06
Feat: sonic lever and button interaction

Mansarde:
## About the PR
This PR adds the ability to toggle a lever via the sonic's _overload_ mode
and the ability to press a stone button via the sonic's _interaction_ mode.

## Why / Balance
It was asked for in here: https://discord.com/channels/1213989169878274068/1402420014350598255
And Saturn was already researching it.

## Technical details
The existing `getHitResult` was insufficient, because it relied on `ProjectileUtil.getCollision`, which only does a raycast that detects blocks it *collides* with.
Since levers and buttons don't have collision, that wouldn't work to detect them.

So I had to add a new method to `SonicMode` called `getHitResultForOutline` that creates a raycast that detects the *outline* of a block instead.

## Media

https://github.com/user-attachments/assets/2a8c30f7-f5d3-4899-bcce-f765fbf9c4d1

## Requirements

**Changelog**
:cl:
- add: Sonic Overload mode can now toggle levers and Sonic Interaction mode can press stone buttons

### DIFF
--- a/src/main/java/dev/amble/ait/core/item/sonic/InteractionSonicMode.java
+++ b/src/main/java/dev/amble/ait/core/item/sonic/InteractionSonicMode.java
@@ -1,5 +1,6 @@
 package dev.amble.ait.core.item.sonic;
 
+import dev.amble.ait.core.AITTags;
 import net.minecraft.block.*;
 import net.minecraft.entity.LivingEntity;
 import net.minecraft.entity.player.PlayerEntity;
@@ -47,6 +48,8 @@ public class InteractionSonicMode extends SonicMode {
     private void interactBlock(BlockPos pos, ServerWorld world, LivingEntity user, int ticks, BlockHitResult blockHit) {
         BlockState state = world.getBlockState(pos);
         Block block = state.getBlock();
+
+        if (!state.isIn(AITTags.Blocks.SONIC_INTERACTABLE)) return;
 
         if (block == Blocks.IRON_DOOR && state.contains(Properties.OPEN)) {
             boolean isOpen = state.get(Properties.OPEN);
--- a/src/main/java/dev/amble/ait/core/item/sonic/InteractionSonicMode.java
+++ b/src/main/java/dev/amble/ait/core/item/sonic/InteractionSonicMode.java
@@ -2,6 +2,7 @@ package dev.amble.ait.core.item.sonic;
 
 import net.minecraft.block.*;
 import net.minecraft.entity.LivingEntity;
+import net.minecraft.entity.player.PlayerEntity;
 import net.minecraft.item.ItemStack;
 import net.minecraft.server.world.ServerWorld;
 import net.minecraft.state.property.Properties;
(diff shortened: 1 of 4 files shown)


src/main/java/dev/amble/ait/core/item/sonic/InteractionSonicMode.java
MaketendoDev, line 1: have you tested that getHitResultForOutline will work for the other interactions?

Mansarde: Yep, overload and interaction both still work with everything the same way it did before the PR. :)

src/main/java/dev/amble/ait/datagen/datagen_providers/AITBlockTagProvider.java
```diff
@@ -53,7 +53,7 @@ protected void configure(RegistryWrapper.WrapperLookup arg) {
                 .add(Blocks.BLACK_CONCRETE,Blocks.CYAN_CONCRETE,Blocks.BLUE_CONCRETE,Blocks.BROWN_CONCRETE,Blocks.GRAY_CONCRETE,Blocks.GREEN_CONCRETE,Blocks.MAGENTA_CONCRETE,Blocks.ORANGE_CONCRETE,Blocks.PINK_CONCRETE,Blocks.RED_CONCRETE,Blocks.WHITE_CONCRETE,Blocks.PURPLE_CONCRETE,Blocks.LIGHT_GRAY_CONCRETE,Blocks.LIGHT_BLUE_CONCRETE,Blocks.LIME_CONCRETE)
                 .add(Blocks.BRICKS, Blocks.BRICK_WALL, Blocks.REDSTONE_TORCH, Blocks.DEEPSLATE_BRICKS)
                 .add(Blocks.BARREL)
-                .add(Blocks.REDSTONE_WIRE, Blocks.COMPARATOR, Blocks.REPEATER)
+                .add(Blocks.REDSTONE_WIRE, Blocks.COMPARATOR, Blocks.REPEATER, Blocks.LEVER)
```
duzos, line 56: if theres a tag for buttons, add them

duzos (changes requested):

src/main/java/dev/amble/ait/datagen/datagen_providers/AITBlockTagProvider.java
Mansarde: Checking for tags is only being done in the Overload mode.
The Interaction mode of the sonic doesn't check for tags and does only block checks:
https://github.com/amblelabs/ait/blob/a685590b4ba40de10eb0e426b059eab29e6d53a7/src/main/java/dev/amble/ait/core/item/sonic/InteractionSonicMode.java#L50-L74

The reason I didn't want to touch them was because I wanted to keep my changes to be additive and not overreach.
So if you want me to check for the button via tags, should I then also convert all the existing checks to look for tags instead as well?

Mansarde: Oh wait, I'm a doofus.
Ignore my previous post.

src/main/java/dev/amble/ait/core/item/sonic/InteractionSonicMode.java
MaketendoDev: okie niceee

PR #1805, opened 2025-08-09
fix console generator breaking in creative mode

Mansarde:
## About the PR
Makes punching (left-clicking) the console generator in creative mode not break it but browse to the previous console.
Breaking the console generator in creative mode (while inside the TARDIS) now only happens when punching it while holding any item in the main hand.
And then it will drop the console generator item, just like when breaking it outside of the TARDIS.

## Why / Balance
To make it consistent with the behavior in survival.
This also prevents accidental breaking while in creative mode.

## Technical details
This PR relies on https://github.com/amblelabs/modkit/pull/35 and uses that to detect whether the player who attempted to break the console generator is sneaking or not.
This information is used to enable the browsing to the *previous* console/variant while in creative mode (see `onTryBreak`).
(Browsing to the *next* console/variant is still handled by `ConsoleGeneratorBlockEntity` and thus doesn't need to be handled by the `onTryBreak`.)

## Media
NOTE: In the videos the sound for browsing to the previous console while in creative mode is missing. This has already been fixed but I didn't record new videos.

https://github.com/user-attachments/assets/7e6f9254-000d-40ec-aed8-b0a19bf2bf22

https://github.com/user-attachments/assets/eeaeb1cf-4ae8-4b3c-8897-5d1d75ddd9a0

## Requirements

**Changelog**
:cl:
- fix: left-clicking breaks the console generator when in creative instead of browsing to the previous console/variant

### DIFF
--- a/.github/workflows/publish-devbuilds.yml
+++ b/.github/workflows/publish-devbuilds.yml
@@ -11,6 +11,7 @@ on:
       - gradlew
       - gradlew.bat
       - versioning.gradle
+      - .github/workflows/publish-devbuilds.yml
 jobs:
   build:
     runs-on: ubuntu-latest
@@ -30,7 +31,7 @@ jobs:
     - name: Build
       run: ./gradlew build
     - name: Publish artifacts
-      uses: DrTheodor/discord-webhook-upload-action@v2.0
+      uses: drtheodor/discord-webhook-upload-action@v0.1
       with:
         url: ${{ secrets.DEV_BUILDS }}
         username: george washington
@@ -39,3 +40,5 @@ jobs:
         message_commit: '> :sparkles: [${commitMessage}](<${commitUrl}>) by [${authorName}](<${authorUrl}>)'
         message_header: |
           <:new1:1253371736510959636><:new2:1253371805734015006> New `Adventures in Time` dev build `#${{ github.run_number }}`:
+
+        file: 'build/libs/*'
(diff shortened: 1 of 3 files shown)


Mansarde:
STATUS: Waiting for next modkit release.

Mansarde:
@duzos For some reason v1.1.13 doesn't show on Jitpack: https://jitpack.io/#amblelabs/modkit
There is only release v1.1.12

Does it just take a while for new releases to show up there, or do I have to do anything special steps to make it appear there?

Mansarde:
Ah, I think you need to make a release on Github itself as well.
At the moment, there is only a 1.1.12 release for modkit: https://github.com/amblelabs/modkit/releases

Mansarde:
Ok, ready for review again. :)

Mansarde:
I reverted my earlier pull of main.
Don't know why I did that, as it added a lot of irrelevant changes/files.
Now this PR is isolated again to the changes it actually touches.

Ready for review (it's a lot less to check know). :)

Mansarde:
Aaaargh, the fix pulled in all kinds of shit from main again.
Gotta fix it...

Mansarde:
All fixed.
Plz review, thx!

MaketendoDev (approved):
Looks good! Tho I would say perhaps maybe change be to a more readable name but its's probably me nitpicking.

Mansarde:
> Looks good! Tho I would say perhaps maybe change be to a more readable name but its's probably me nitpicking.

I just copied how it was named in the `ConsoleGeneratorBlock` class, to keep it consistent.
I'll leave it for now (as the name is defined just a few lines above its use), but if I touch the console generator classes again in the future I'll rename them all. :)

Addi3:
btw in creative mode, i feel like the console generator should not drop when broken

PR #1824, opened 2025-08-15
Gradle build config: Make multidim, scheduler and queue libraries transitive

Mansarde:
## About the PR
Make multidim, scheduler and queue libraries transitive.

## Why / Balance
If a mod uses AIT as a dependency (like e.g. ait-extras), then these requried libraries will automatically be available as well without having to be explicitly included.

## Technical details
Normally `modApi` is transitive, but when used within an `include` then the non-transitivity of that wins.
In order to make it transitive again I simply added a `transitive = true` to each include block.

## Requirements

**Changelog**
:cl:
- chore: make multidim, scheduler and queue libraries transitive

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -85,12 +85,15 @@ dependencies {
     // TODO: move these 3 to modkit so we don't have to re-export so many dependencies
     include(modApi("dev.drtheo:multidim:${project.multidim_version}")) {
         exclude(group: "net.fabricmc.fabric-api")
+        transitive = true
     }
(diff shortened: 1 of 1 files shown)


drtheodor (approved):
LGTM, needs testing

Mansarde:
Thx, I tested it on a local build of ait-extras and it seemed to work.
I have a commit ready for ait-extras that was just waiting for this to merge.
Will test that now after the merge.

duzos:
do it for mod kit too pls

PR #1829, opened 2025-08-17
Feat: Add engine overdrive arming mechanism

Mansarde:
## About the PR
This PR adds an arming mechanism to the engine overdrive.
                                                               
First press arms it.
Pressing again within 5 seconds executes overdrive.
Pressing any other control or waiting 5 seconds disarms it.
The 5 minute cooldown only starts after overdrive was executed.

## Why / Balance
To prevent accidental activation.

## Technical details
I added two static functions to the engine overdrive control that arm and disarm it:
`arm()` sets a boolean to true and starts a confirmation-timer that will disarm the overdrive after 5 seconds (20 * 5 ticks).
`disarm()` sets the same boolean to false and cancels the confirmation-timer.

When the control is first pressed, overdrive will be armed and return.
If the control is pressed again within the confirmation-time, overdrive is disarmed (just to reset the state) and the dumping of Artron is allowed to commence.

To cancel the event whenever another control is pressed (after arming overdrive):
I added an event that fires when a control is used.
The engine overdrive control registers to this event and if any other control is used it disarms overdrive.

## Requirements

**Changelog**
:cl:
- add: arming mechanism for engine overdrive

### DIFF
--- a/src/main/java/dev/amble/ait/api/tardis/TardisEvents.java
+++ b/src/main/java/dev/amble/ait/api/tardis/TardisEvents.java
@@ -288,9 +288,9 @@ public final class TardisEvents {
                 }
             });
     public static final Event<UseControl> USE_CONTROL = EventFactory.createArrayBacked(UseControl.class,
-            callbacks -> control -> {
+            callbacks -> (control, tardis, player, world, console, leftClick) -> {
                 for (UseControl callback : callbacks) {
-                    callback.onUse(control);
+                    callback.onUse(control, tardis, player, world, console, leftClick);
                 }
             });
 
@@ -550,7 +550,7 @@ public final class TardisEvents {
          * @param control
          *            the control that was used
          */
-        void onUse(Control control);
+        void onUse(Control control, Tardis tardis, ServerPlayerEntity player, ServerWorld world, BlockPos console, boolean leftClick);
     }
 
 
--- a/src/main/java/dev/amble/ait/api/tardis/TardisEvents.java
+++ b/src/main/java/dev/amble/ait/api/tardis/TardisEvents.java
@@ -2,6 +2,7 @@ package dev.amble.ait.api.tardis;
 
 import java.util.Optional;
 
+import dev.amble.ait.core.tardis.control.Control;
 import dev.amble.lib.data.CachedDirectedGlobalPos;
 import dev.amble.lib.data.DirectedBlockPos;
 import net.fabricmc.fabric.api.event.Event;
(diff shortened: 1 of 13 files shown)


Addi3:
peak

src/main/java/dev/amble/ait/api/tardis/TardisEvents.java
```diff
@@ -535,6 +542,17 @@ public interface OnEnginesPhase {
...
 
+    @FunctionalInterface
+    public interface UseControl {
+        /**
+         * Called when a console control is used.
+         *
+         * @param control
+         *            the control that was used
+         */
+        void onUse(Control control);
```
drtheodor, line 553: this should, ideally, have all the same arguments as the ones the real control gets on use, that is being:
- Tardis
- ServerPlayerEntity
- ServerWorld 
- BlockPos of the console
- boolean leftClick

src/main/java/dev/amble/ait/core/tardis/control/impl/EngineOverloadControl.java
```diff
@@ -26,6 +29,17 @@ public class EngineOverloadControl extends Control {
 
     private static final Random RANDOM = new Random();
     private static final String[] SPINNER = {"/", "-", "\\", "|"};
+    private static final long CONFIRMATION_TIME = 20 * 5;   //in ticks
+    private static Task confirmationTimer;
+    private static boolean isArmed = false;
```
drtheodor, line 34: uh oh! state in stateless code!

drtheodor (changes requested):
controls don't have state, you should instead delegate all this server-side logic from the control to a tardis handler or get rid of the state completely

src/main/java/dev/amble/ait/core/tardis/control/impl/EngineOverloadControl.java
Mansarde: I mean, the state is only temporary and resets after 5 seconds or when another control was used.
So it's not really a state that would need to be persisted or anything.
Is it really bad in this case?

drtheodor: controls objects are not per-tardis/console
they are global

drtheodor: so when you add a 5 seconds timer to it, it will "add it to all engine overload controls"

drtheodor (approved):

PR #1838, opened 2025-08-22
Build performance improvements

drtheodor:
## About the PR
This PR aims to improve performance of the mod by decreasing its file size (in release) and making build times faster in development.

## Why / Balance
The mod is growing bigger by the day. This PR aims to fix this by improving the build times in development and decreasing file size in release. As things should be.
As a result, the build went from `55.28 MB` to `26.42 MB`.

## Technical details
CI will build the mod, then use PackSquash for the resources at `build/resources/main/` (which also includes the generated resources at this point), which generates a zip file in the temp folder. Then, a bash script called `merge_jars` gets ran, which unzips the built jar into a temp folder and then unzips the generated squashed resources into the very same folder with overwrite. This is done because 
1. PackSquash doesn't support squashing zips directly.
2. Using PackSquash directly on unzipped jar contents bloats PackSquash's config and worsens the build times.

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: the mod is now almost twice as smaller! yippe!


Mansarde (commented):
Do all the files in `src/main/generated/.cache` need to be tracked in the repo?
Or even the entire `src/main/generated` folder itself?

drtheodor:
> Do all the files in `src/main/generated/.cache` need to be tracked in the repo? Or even the entire `src/main/generated` folder itself?

yes

Mansarde (approved):
LGTM

PR #1839, opened 2025-08-24
feat: sonic mode check for phasing

rapbattlegod32:
## About the PR
[](feat: sonic mode check for phasing)

## Why / Balance
[](feat: sonic mode check for phasing)

## Technical details[
](feat: sonic mode check for phasing)

## Requirements

:cl:
- feat: check for sonic mode tardis to repair phasing engines

### DIFF
--- a/src/main/java/dev/amble/ait/core/blockentities/ExteriorBlockEntity.java
+++ b/src/main/java/dev/amble/ait/core/blockentities/ExteriorBlockEntity.java
@@ -4,6 +4,7 @@ import static dev.amble.ait.core.tardis.handler.InteriorChangingHandler.MAX_PLAS
 
 import java.util.UUID;
 
+import dev.amble.ait.core.item.sonic.SonicMode;
 import dev.amble.lib.data.CachedDirectedGlobalPos;
 import dev.drtheo.scheduler.api.TimeUnit;
 import dev.drtheo.scheduler.api.common.Scheduler;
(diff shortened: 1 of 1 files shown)


src/main/java/dev/amble/ait/core/blockentities/ExteriorBlockEntity.java
```diff
@@ -163,8 +164,10 @@ public void useOn(ServerWorld world, boolean sneaking, PlayerEntity player) {
             EngineSystem.Phaser phasing = tardis.subsystems().engine().phaser();
 
             if (phasing.isPhasing()) {
-                world.playSound(null, pos, AITSounds.SONIC_USE, SoundCategory.PLAYERS, 1F, 1F);
-                phasing.cancel();
+                if(SonicItem.mode(hand) == SonicMode.Modes.TARDIS) {
+                    world.playSound(null, pos, AITSounds.SONIC_USE, SoundCategory.PLAYERS, 1F, 1F);
+                    phasing.cancel();
+                }
```
Mansarde, line 170: I would combine the conditions, like so:

```suggestion
if (phasing.isPhasing() && SonicItem.mode(hand) == SonicMode.Modes.TARDIS) {
                world.playSound(null, pos, AITSounds.SONIC_USE, SoundCategory.PLAYERS, 1F, 1F);
                phasing.cancel();
```

The first condition doesn't need to be separate for the return statement, because if we don't use the sonic in TARDIS mode then it should be just like as if we hadn't used the sonic to cancel the phasing, so it should actually continue processing the rest of the method.

Otherwise, while holding the sonic in any other mode while phasing is happening, then we wouldn't be able to interact with the doors of the TARDIS (i.e. the `tardis.door().interact()` further down).

Mansarde: I don't know why the *before* part of the suggestion has both codes in it, just ignore that and look at the *after* part. xD

rapbattlegod32: alr thank  you ill do this when i get home

rapbattlegod32:
@Mansarde can you review please

Mansarde (approved):
LGTM

PR #1852, opened 2025-08-27
Fix/power converter bugs

### LINKED ISSUE #1849: use the power converter by right clicking it with a empty hand
### What happened?

get a power converter
connect it
right click it with a empty hand
free fuel

### Version

1.2.11

### Mod Loader

fabric

### Build number & branch

_No response_

### Logs & screenshots

_No response_

Mansarde:
## About the PR
This PR fixes:
- the Power Converter being fueled by an empty hand.
- the Power Converter instantly draining all Artron when adding fuel during power off.

fixes #1849

## Why / Balance
An empty hand shouldn't count as fuel and the Artron reserves shouldn't vanish when adding fuel via the power converter during power off.

## Technical details
<ins>Regarding "being fueled by an ampty hand":</ins>
Before this PR the TARDIS_CORAL_WALL_FAN was registered as a fuel item.
TARDIS_CORAL_WALL_FAN is a `@NoBlockItem` item though, which means no actual item will be created for it.
From what I could find, it doesn't spawn naturally, nor is there a recipe for it.
Despite this it was added as a possible fuel source.

The problem is:
Due to the `@NoBlockItem`, this item will be registered as `minecraft:air` (during datagen).
The result is that right-clicking e.g. the Power Converter with an empty hand will make it think that we just gave it fuel, because an empty hand apparently counts as holding `minecraft:air`.

Since the TARDIS_CORAL_WALL_FAN cannot be an item in-game anyways and thus cannot be fed to the Power Converter, it makes no sense to count it as fuel.
By removing it as a fuel item, this solves both problems.

<ins>Regarding the instant draining of all Artron by the power converter:</ins>
This is where there power converter adds fuel:
https://github.com/amblelabs/ait/blob/2fb99f547973732d413b9540cded6347c5d042b6/src/main/java/dev/amble/ait/core/blocks/PowerConverterBlock.java#L76-L84

When the TARDIS is powered off, then `be.source()` will return `0.0` instead of the actual Artron level.
So as soon as someone then adds fuel to the power converter, the fuel level will be *set* to `0.0 + 175` when not sneaking or `0.0 + 175 * count` (where `count` could be 0, thus resulting in `0.0`) when sneaking.

## Requirements

## Breaking changes

**Changelog**
:cl:
- fix: Power Converter could be fueled by an empty hand
- fix: Power Converter would instantly drain all Artron when trying to fuel it during power off

### DIFF
--- a/src/main/java/dev/amble/ait/core/engine/link/ITardisSource.java
+++ b/src/main/java/dev/amble/ait/core/engine/link/ITardisSource.java
@@ -8,7 +8,6 @@ public interface ITardisSource extends IFluidSource {
     @Override
     default double level() {
         if (getTardisForFluid() == null) return 0;
-        if (!getTardisForFluid().fuel().hasPower()) return 0;
 
         return getTardisForFluid().fuel().getCurrentFuel();
     }


rapbattlegod32:
did you fix the bug where you have to replace the power convereter for it to work, (place power convereter and leave and reenter tardis for it to not work)

Mansarde:
> did you fix the bug where you have to replace the power convereter for it to work, (place power convereter and leave and reenter tardis for it to not work)

No, that's a separate issue. Haven't looked into that one.

PR #1856, opened 2025-08-28
feat: handles improvements

### LINKED ISSUE #1751: Handles Upgrade
### Description

Add the sonic flight events indicator functionality to Handles when plugged into the console unit. 

.

rapbattlegod32:
## About the PR
handles can now be used in the sonic port to show the flight sequence events (indication thingy) like the sonic scanning mode, and im also going to implement the levenshtein distance thingy theo requested to make handles more retard proof

## Why / Balance
closes #1751 

## Technical details
new method in controlentityrenderer for seeing if handles are in a console

## Media
image

## Requirements

**Changelog**

:cl:
- feat: Added handles flight indicators when in sonic port!
- tweak: Added levenshtein distance when calling handles!


src/main/java/dev/amble/ait/client/renderers/entities/ControlEntityRenderer.java
```diff
@@ -173,6 +173,17 @@ private static boolean isScanningSonicInConsole(ConsoleControlEntity entity) {
         return SonicRendering.isScanningSonic(sonic);
     }
 
+    private static boolean isHandlesInConsole(ConsoleControlEntity entity) {
+        if (entity.getConsole() == null) return false;
+
+        ConsoleBlockEntity console = entity.getConsole();
```
Mansarde, line 179: Put this above the guard clause, so you don't have to call `entity.getConsole()` two times, like so:
```java
ConsoleBlockEntity console = entity.getConsole();
if (console  == null) return false;
```

src/main/java/dev/amble/ait/client/renderers/entities/ControlEntityRenderer.java
```diff
@@ -173,6 +173,17 @@ private static boolean isScanningSonicInConsole(ConsoleControlEntity entity) {
...
+
+        if (!console.isLinked()) return false;
+
+        Tardis tardis = console.tardis().get();
```
Mansarde, line 183: I'm not sure if it's necessary, but I'd probably add a null check for the Tardis object.
Perhaps you can add it to the return statement, like so:
```java
return (tardis != null && tardis.butler().getHandles() != null)
```

Mansarde (changes requested):

src/main/java/dev/amble/ait/client/renderers/entities/ControlEntityRenderer.java
```diff
@@ -173,6 +173,16 @@ private static boolean isScanningSonicInConsole(ConsoleControlEntity entity) {
...
     }
 
+    private static boolean isHandlesInConsole(ConsoleControlEntity entity) {
+        ConsoleBlockEntity console = entity.getConsole();
+        if (console  == null) return false;
+
+        if (!console.isLinked()) return false;
+
+        Tardis tardis = console.tardis().get();
+        return (tardis != null && tardis.butler().getHandles() != null)
```
Mansarde, line 183: Sorry for this little nitpick, but this line needs a `;` at the end, otherwise it won't compile. 😅

Mansarde (changes requested):

Mansarde:
In terms of the flight indicators this looks good now and I also tested it succesfully in-game. 👍
I can see that you also want to improve Handle's command interpretation, so I'll hold off with approvals until that's done.

rapbattlegod32:
ok ive done it but ima keep it a buck and be honest i dont know how half of this code works but it works

src/main/java/dev/amble/ait/client/config/AITClientConfig.java
```diff
@@ -83,6 +83,10 @@ public class AITClientConfig {
     @Boolean(formatter = Boolean.Formatter.YES_NO, colored = true)
     @SerialEntry public boolean animateDoors = true;
 
+    @AutoGen(category = CATEGORY)
+    @IntField(min = 0)
+    @SerialEntry public int handlesLevenshteinDistance = 2;
```
MaketendoDev, line 88: Upload an image and a description for the config (look in ait-polishin branch)

src/main/java/dev/amble/ait/core/handles/HandlesResponse.java
```diff
@@ -69,6 +69,34 @@ default boolean isCommand(String command) {
         return getCommandWords().contains(command.toLowerCase());
     }
 
+    // i have no idea how this shit fucking works
+    default int distance(String command, String input) {
```
MaketendoDev, line 73: I feel like theo would absolutely kill you from this?

MaketendoDev (commented):
I think the code is good just show theo your distance code to see how he likes it and then fix the config thing

src/main/java/dev/amble/ait/core/handles/HandlesResponse.java
```diff
@@ -69,6 +69,34 @@ default boolean isCommand(String command) {
...
+        int[][] dp = new int[command.length() + 1][input.length() + 1];
```
Mansarde, line 74: Wouldn't it be enough to initiate the array with the exact lengths instead of length + 1?
Because nowhere in the method does it ever access the length + 1 element.

For example, if command was `refuel`, which has a length of 6, then the loop in the method would iterate from 1 to 6 (both including).
Since we never access index 7 (which is length + 1) anywhere, it should suffice to use only the length to initialize the array (for both command and input).

```suggestion
        int[][] dp = new int[command.length()][input.length()];
```

Mansarde (changes requested):

Mansarde:
Nevermind my last comment, it's invalid.

src/main/java/dev/amble/ait/core/handles/HandlesResponse.java
```diff
@@ -69,6 +69,34 @@ default boolean isCommand(String command) {
...
+
+                dp[i][j] = Math.min(
+                        Math.min(
+                                dp[i - 1][j] + 1,
+                                dp[i][j - 1] + 1
+                        ),
+                        dp[i - 1][j - 1] + cost
+                );
+            }
+        }
```
Mansarde, line 95: The algorithm looks good, but instead of using two separate loops to set all `dp[0][j]` and all `dp[i][0]`, you can reduce the number of loops by making them part of the main loop, like so:

```suggestion
        for (int i = 1; i <= command.length(); i++) {
            for (int j = 1; j <= input.length(); j++) {
                if (i == 0) {
                    dp[i][j] = j;
                } else if (j == 0) {
                    dp[i][j] = i;
                }
                
                int cost = (command.charAt(i - 1) == input.charAt(j - 1)) ? 0 : 1;

                dp[i][j] = Math.min(
                        Math.min(
                                dp[i - 1][j] + 1,
                                dp[i][j - 1] + 1
                        ),
                        dp[i - 1][j - 1] + cost
                );
            }
        }
```

src/main/java/dev/amble/ait/core/handles/HandlesResponse.java
rapbattlegod32: i watched like different videos on it and one with a realy strong japanese acent bro i was going to kill myself

drtheodor: Make the levenshtein distance a util method

src/main/java/dev/amble/ait/client/config/AITClientConfig.java
```diff
@@ -83,6 +83,12 @@ public class AITClientConfig {
     @Boolean(formatter = Boolean.Formatter.YES_NO, colored = true)
     @SerialEntry public boolean animateDoors = true;
 
+    @AutoGen(category = CATEGORY)
+    @IntField(min = 0)
+    @CustomDescription(value = "The levenshtein distance allows for typos when using handles. Distances lower to 0 are stricter while higher values like 5 are more lenient.")
+    @CustomImage(value = "textures/yacl3/server/levenshetin.png", width = 1909, height = 349)
```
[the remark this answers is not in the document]
MaketendoDev: and it needs to be webp

duzos:
@rapbattlegod32 bump

PR #1857, opened 2025-08-28
Fix celsius showing fahrenheit degrees

Mansarde:
## About the PR
Makes the space suit display Celsius degrees when Celsius setting is chosen instead of displaying Fahrenheit degrees.
Also fixes some Celsius spelling and improves the temperature config enum code.

## Why / Balance
Because Celsius is not Fahrenheit.

## Technical details
Changed a call to `celsius()` instead of `fahrenheit()` where Celsius was called for (pun intended).

## Requirements

**Changelog**
:cl:
- fix: Celsius was showing Fahrenheit degrees

### DIFF
--- a/src/main/java/dev/amble/ait/client/config/AITClientConfig.java
+++ b/src/main/java/dev/amble/ait/client/config/AITClientConfig.java
@@ -12,6 +12,7 @@ import dev.isxander.yacl3.platform.YACLPlatform;
 import net.minecraft.text.Text;
 
 import dev.amble.ait.AITMod;
+import org.apache.commons.lang3.StringUtils;
 
 public class AITClientConfig {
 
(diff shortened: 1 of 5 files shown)


src/main/java/dev/amble/ait/client/config/AITClientConfig.java
```diff
@@ -94,17 +95,12 @@ public enum TemperatureType implements NameableEnum {
...
-            key1 = switch (key1) {
-                case "Celsius" -> key1 + " (°C)";
-                case "Fahrenheit" -> key1 + " (°F)";
-                case "Kelvin" -> key1 + " (K)";
-                default -> key1;
+            String unitSymbol = switch (this) {
+                case CELSIUS -> "°C";
+                case FAHRENHEIT -> "°F";
+                case KELVIN -> "K";
             };
```
drtheodor, line 102:
```suggestion
```

src/main/java/dev/amble/ait/client/config/AITClientConfig.java
```diff
@@ -94,17 +95,12 @@ public enum TemperatureType implements NameableEnum {
...
-            this.key = key1;
+            this.key = String.format("%s (%s)", StringUtils.capitalize(this.toString().toLowerCase()), unitSymbol);
```
drtheodor, line 103:
```suggestion
            this.key = this.key = "text.config.aitconfig.enum.temperatureType." + this.toString().toLowerCase();
```

this should be a translation key, not the actual value!

drtheodor (changes requested):

src/main/java/dev/amble/ait/client/config/AITClientConfig.java
Mansarde: Hmm, these two look the same to me. 🤔

PR #1866, opened 2025-09-02
Fix explosion (and fire) gamerules

Mansarde:
## About the PR
Fixes the gamerules `tardisGriefing` and `tardisFireGriefing`.
If `tardisGriefing` is set to `true`, then explosions will destroy blocks, otherwise not.
If `tardisFireGriefing` is set to `true`, then explosions will also create fire (only if `tardisGriefing` is also true), otherwise not.

## Why / Balance
The gamerules were only doing something effective for a crashing TARDIS that was falling.
No other explosions (of *any* fires) were controlled by them.

## Technical details
I extracted the `ExplosionBehavior` from `FallingTardisEntity` and put it in `TardisUtil` so it can be used from one central location.
Then I added this ExplosionBEhavior to all other explosions.
Additionally I added a `doCreateFire` method to `TardisUtil` that checks the gamerule `tardisFireGriefing` and added a call to it to all the explosions as well.

<ins>**NOTE:**</ins>
I didn't touch the explosions of the "avoid debris" flight event (in [`SequenceRegistry`](https://github.com/amblelabs/ait/blob/main/src/main/java/dev/amble/ait/registry/impl/SequenceRegistry.java)), because those explosions only act as a damage-source to hurt the player (i.e. they are *never* meant to destroy blocks, regardless of gamerule).

## Requirements

**Changelog**
:cl:
- fix: Gamerules did not control all explosions (and their fire creation)

### DIFF
--- a/src/main/java/dev/amble/ait/core/blocks/PeanutBlock.java
+++ b/src/main/java/dev/amble/ait/core/blocks/PeanutBlock.java
@@ -1,10 +1,12 @@
 package dev.amble.ait.core.blocks;
 
+import dev.amble.ait.core.tardis.util.TardisUtil;
 import net.minecraft.block.Block;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.world.World;
 
 public class PeanutBlock extends Block {
+    private static final float EXPLOSION_POWER = 100;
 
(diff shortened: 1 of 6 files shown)


drtheodor (approved):
LGTM

PR #1867, opened 2025-09-05
fix crashing still allowing to land in locked dimensions

Mansarde:
## About the PR
I completely forgot about the crashing part in #1836 
So this PR fixes that, i.e. crashing will now definitely not allow landing in locked dimensions.

## Why / Balance
I previously only fixed it for landings with Ghost Monument disabled.
This will ensure that crashing only happens into unlocked dimensions.

## Technical details
Crashing will revert the destination world to the previous world before landing.

## Requirements

**Changelog**
I'm not adding a changelog here, because the changelog from #1836 already covers this PR.
(Unless there's a good reason to add it of course. Let me know!)


duzos (approved):
LGTM

PR #1876, opened 2025-09-08
fix waterlogged not being removed from all doors

Mansarde:
## About the PR
It removes the waterlogged status from *all* interior doors instead of just from the last placed/interacted door.

## Why / Balance
In #1738 I only removed the state from the last placed/interacted door, which will leave all other doors waterlogged and continue to reduce loyalty, even when the doors were closed or the shields activated.
But in that case *all* the doors should no longer be waterlogged.

## Technical details
I moved the removal logic for the waterlogged status from the door handler to the doors themselves.
This way we don't need to save the positions for all the doors and can just let each door handle their own waterlogged status based on whether it is closed or the shields are active.

## Requirements

**Changelog**
:cl:
- fix: When an interior door was closed or the shields activated, only the last placed/interacted door would no longer be waterlogged, all the others would remain waterlogged (and continue to reduce loyalty).

### DIFF
--- a/src/main/java/dev/amble/ait/core/blockentities/DoorBlockEntity.java
+++ b/src/main/java/dev/amble/ait/core/blockentities/DoorBlockEntity.java
@@ -61,7 +61,7 @@ public class DoorBlockEntity extends InteriorLinkableBlockEntity {
 
         Tardis tardis = door.tardis().get();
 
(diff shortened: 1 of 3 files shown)


rapbattlegod32 (approved):
lgtm

PR #1888, opened 2025-09-12
fix some chunk deadlocks

drtheodor:
## About the PR
This PR fixes some chunk deadlocks.

## Technical details
By offsetting the post init to be a tick later this *should* eliminate the problems with deadlocks.

## Requirements

**Changelog**

<!--
:cl:
- fix: no more chunk deadlocks with a lot of optimization mods


Mansarde (approved):
I think understand the flow of the processing, though I couldn't say if there are any issues or not.
It looks like you make sure that the postInit is run after a tick, to give the init enough time to process.
If my understanding of this is good enough for you, then this LGTM. 😁

PR #1902, opened 2025-09-19
restore NONE for vertical search and disable falling directly instead

Mansarde:
## About the PR
Restore landing (and hovering) in air, but keep falling disabled.

Since this is only changing a temporary hotfix to another temporary hotfix, I don't add a changelog.

## Why / Balance
NONE is temporarily disabled as an option due to issues with falling TARDISes.
This restores the option again and instead makes TARDISes not fall, so that the TARDIS can still be landed in the air.

## Technical details
`NONE` is restored for vertical search.
On falling, I just return instead of processing the fall.

## Requirements

## Breaking changes

**Changelog**
:nocl:

### DIFF
--- a/src/main/java/dev/amble/ait/core/tardis/handler/RealFlightHandler.java
+++ b/src/main/java/dev/amble/ait/core/tardis/handler/RealFlightHandler.java
@@ -95,6 +95,7 @@ public class RealFlightHandler extends KeyedTardisComponent implements TardisTic
     }
 
     public void onStartFalling(ServerWorld world, BlockState state, BlockPos pos) {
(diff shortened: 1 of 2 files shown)


Mansarde:
> Looks alright if it wasn't for the fact that we're doing useless block checks now... The `if (true) return;` should ideally be called in the block entity before it does the block state checks and stuff.

Does that really matter if this is just a temporary measure?
I'll change it though, no problem. And to be honest, using `if (true)` always feels just wrong to me, so it's probably best to not have it in there, even if temporary.
Who knows how many times people thought something was temporary, just for it to persist over months or years and then become forgotten. 😅

Mansarde:
Ok, pushed a new commit that comments out the `onStartFalling` call instead. :)

PR #1913, opened 2025-09-26
Fix siege cubes remaining in player inventory on death via keepinventory

Mansarde:
## About the PR
Prevents siege cubes from staying in a player's inventory on death when the gamerule `keepinventory` is true.

Originally reported by `amokdev` on Discord here: https://discord.com/channels/1213989169878274068/1421125485362020525/1421134182800818206

## Why / Balance
So that a player, who set their spawn point inside a TARDIS, cannot take the siege cube inside of that TARDIS by dying with the siege cube in their inventory (while keepinventory is true).

## Technical details
I hooked into the `onDeath()` method of the `ServerPlayer` and then check for siege cube items.
If one is found, I remove it and place the TARDIS exterior at the position of the player.

That code is basically just copied over from how it is done when a player is disconnecting from the server with a siege cube in their inventory (see `ServerPlayConnectionEvents.DISCONNECT` event)

## Requirements

**Changelog**
:cl:
- fix: Siege cubes remained in a player's inventory on death when the gamerule "keepinventory" was true

### DIFF
--- a/src/main/java/dev/amble/ait/mixin/server/ServerPlayerMixin.java
+++ b/src/main/java/dev/amble/ait/mixin/server/ServerPlayerMixin.java
@@ -1,5 +1,11 @@
 package dev.amble.ait.mixin.server;
 
+import dev.amble.ait.core.AITItems;
+import dev.amble.ait.core.item.SiegeTardisItem;
+import dev.amble.ait.core.tardis.manager.ServerTardisManager;
+import net.minecraft.entity.damage.DamageSource;
+import net.minecraft.item.ItemStack;
+import net.minecraft.item.Items;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Inject;
@@ -10,6 +16,8 @@ import net.minecraft.server.network.ServerPlayerEntity;
 import dev.amble.ait.core.entities.FlightTardisEntity;
 import dev.amble.ait.core.tardis.util.TardisUtil;
 import dev.amble.ait.core.world.TardisServerWorld;
+
+import java.util.Objects;
 
 @Mixin(ServerPlayerEntity.class)
 public class ServerPlayerMixin {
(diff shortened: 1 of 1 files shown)


src/main/java/dev/amble/ait/mixin/server/ServerPlayerMixin.java
```diff
@@ -31,4 +39,26 @@ public class ServerPlayerMixin {
         if (player.hasVehicle() && player.getVehicle() instanceof FlightTardisEntity)
             ci.cancel();
     }
+
+    @Inject(method = "onDeath", at = @At("HEAD"))
+    public void ait$onDeath(DamageSource damageSource, CallbackInfo ci) {
+        ServerPlayerEntity player = (ServerPlayerEntity) (Object) this;
+
+        ServerTardisManager.getInstance().forEach(tardis -> {
```
drtheodor, line 47: ow. this is real bad.

src/main/java/dev/amble/ait/mixin/server/ServerPlayerMixin.java
```diff
@@ -31,4 +39,26 @@ public class ServerPlayerMixin {
...
+            if (!tardis.siege().isActive())
+                return;
+
+            if (!Objects.equals(tardis.siege().getHeldPlayerUUID(), player.getUuid()))
+                return;
+
+            for (ItemStack itemStack : player.getInventory().main) {
```
drtheodor, line 54: thats horrible

drtheodor (changes requested):

Mansarde:
> just prevent dropping siege cubes inside the tardis

I thought about this first, but I assumed that if someone's game crashes while they have a cube in their inventory then there would be no exterior in the world anymore (not even in intem form).

But I just tested in ingame and it seems the disconnect event ( with the *real bad* code :P ) is actually preventing that from happening as well.

So I'll just prevent the dropping of the item in the same TARDIS then as you suggested. 🫡

Mansarde:
Hmm, this problem is a bit more extensive than I initially thought.
For example, I have to also consider what happens if a player dies inside the TARDIS they have a siege cube of in their inventory.
Or that a player could put the item inside a container that is inside the sieged TARDIS.
Or the various ways in which an item can be dropped from the inventory.
Etc.

Will have to analyze this a bit more to cover all the bases.

src/main/java/dev/amble/ait/mixin/server/ServerPlayerMixin.java
Mansarde: So you did the same thing as I did then. Copy it without changing it. :P

But jokes aside, I guess the point is moot anyways, now that this is not going to be used at all.

drtheodor:
https://discord.com/channels/1213989169878274068/1213989170964340883/1421197555156389981

PR #1915, opened 2025-09-26
Feat: Evict players who fall into void without a key when leave-behind is on

Mansarde:
## About the PR
When leave-behind is enabled and a player falls into the TARDIS void, then without a key (and valid loyalty level) they will be spawned outside the TARDIS instead of at the interior door.

## Why / Balance
This was a a suggested idea on Discord by `ihudhudi` here:
https://discord.com/channels/1213989169878274068/1420864949474951189/1420864949474951189

## Technical details
Added an additional check to the existing ServerPlayerMixin to check for a matching key and the leave-behind setting.

## Media
image

## Requirements

**Changelog**
:cl:
- add: When leave-behind is on, then players falling into the TARDIS void without a key (and valid loyalty level) will be evicted from the TARDIS.


drtheodor (approved):
LGTM

PR #1922, opened 2025-10-07
fix: smoother vortex animation

duzos:


src/main/java/dev/amble/ait/client/renderers/VortexRender.java
```diff
@@ -37,50 +35,64 @@ public VortexUtil(Identifier texture) {
...
+    public boolean isFor(Identifier texture) {
+        return this.texture.equals(texture);
+    }
 
-        MinecraftClient.getInstance().getTextureManager().bindTexture(TEXTURE_LOCATION);
-        Tessellator tessellator = Tessellator.getInstance();
-        BufferBuilder buffer = tessellator.getBuffer();
+    public void replaceWith(Identifier texture) {
+        this.texture = texture;
+        secondLayer = new Identifier(texture.getNamespace(), texture.getPath().substring(0, texture.getPath().length() - 4) +
```
drtheodor, line 68: why dont you pass the vortex name here?

drtheodor (changes requested):

PR #1933, opened 2025-10-18
Fix/painting-drop

lakehascake:
## About the PR
- Made Paintings actually drop their respective items when broken.

## Why / Balance
- This is how paintings are supposed to work in game.
- This was requested by players to be fixed.

## Technical details
- Made BOTIPaintingEntity abstract
- General clean-up after previous change to make sure the code functions as intended

## Media
N/A

## Requirements

- [x ] I have read and am following the [Pull Request and Changelog Guidelines](https://amblelabs.github.io/ait-wiki/guidelines).
- [x ] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes
N/A

**Changelog**
:cl:
- fix: Fixed the BOTI paintings not dropping items when broken in survival.

### DIFF
--- a/src/main/java/dev/amble/ait/core/entities/BOTIPaintingEntity.java
+++ b/src/main/java/dev/amble/ait/core/entities/BOTIPaintingEntity.java
@@ -31,7 +31,7 @@ public abstract class BOTIPaintingEntity extends AbstractDecorationEntity implem
     public static Optional<BOTIPaintingEntity> placePainting(EntityType<? extends BOTIPaintingEntity> entityType, World world, BlockPos pos, Direction facing) {
         BOTIPaintingEntity paintingEntity = entityType.create(world);
 
-        assert paintingEntity != null;
+        if (paintingEntity == null) return Optional.empty();
         paintingEntity.setPosition(pos.getX(), pos.getY(), pos.getZ());
         paintingEntity.setFacing(facing);
 


src/main/java/dev/amble/ait/core/entities/BOTIPaintingEntity.java
```diff
@@ -19,21 +19,20 @@
...
 
-    private BOTIPaintingEntity(EntityType<? extends BOTIPaintingEntity> entityType, World world, BlockPos pos) {
-        super(entityType, world, pos);
-    }
 
     public static Optional<BOTIPaintingEntity> placePainting(EntityType<? extends BOTIPaintingEntity> entityType, World world, BlockPos pos, Direction facing) {
-        BOTIPaintingEntity paintingEntity = new BOTIPaintingEntity(entityType, world, pos);
+        BOTIPaintingEntity paintingEntity = entityType.create(world);
 
+        assert paintingEntity != null;
```
Mansarde: `assert` shouldn't be used outside of testing.
I would recommend a proper null check instead and perhaps return `Optional.empty()` in case of null.

lakehascake: that was there when i got into the code, but i'll change it rq

lakehascake: wait no it wasnt nvm one sec

lakehascake: fixed in most recent commit, thank you for your help

Mansarde (approved):
LGTM

Issue #1381, opened 2025-04-04
H.A.D.S


Addi3:
### Description

(ik technically we have it but its literally just the alarms sooo)

### Two types of H.A.D.S (maybe. they are both activated the same way but just have different results)

- Type 1: Hostile Action Displacement System:
Basically the tardis moves to a random location (50-100 blocks) from the current point of the tardis when hads are triggered.

- Type 2: Hostile Action Dispersal System: 
Instead of type 1, type 2 (when triggered) will result in the tardis 'exploding' into particles (could use the ait particles we have rn for the engine and growth) and it will lock and trigger cloak. Sonicing the tardis will uncloak and unlock it (or the player will have to manually do both 🤷) and the same particles will go towards the middle of thae exterior block.

### Trigger events
A list of things that will trigger H.A.D.S:

- projectiles hitting the tardis (arrows, tridents, staser bullets)
- breaking open the doors / knocking on them when locked
- landing in lava
- landing near "hostile zones" (areas with a ton of mobs)
- landing in disliked areas (links with this >https://github.com/amblelabs/ait/issues/908)
- (feel free to suggest more stuff i didnt think off)

### Side note:

For balance it can damage a circuit or multiple maybe 

### Images, links & related resources

- HADS TYPE 1 > https://www.youtube.com/watch?v=MEnQ6ETBxhE
- HADS TYPE 2 > https://youtu.be/Hu1NpY6-PRw?t=30 (this is the reconstruction cus i couldnt find the deconstruction scene. is S9 E2)

### I've checked the feature requests...

- [x] ...and this feature was not requested before.

Saturnorsomthing:
i love the 2nd type

TokenBro11400:
idea: since the tardis kinda goes poof it can lock the doors for anyone inside or maybe its just a door on the other side since the exterior isnt whole aswell as making the hitbox so you can walk through it. and possibly a mark like the summoning thing while holding the sonic to show where to stand to be sent inside once reconstructed. other idea: have a small animation or screen effect for the tardis materializing/reconstructing on you aswell as a security setting to allow people without keys get teleported in when materializing/reconstructing on them

DrTheodor:
There's a HADS code present in the codebase just... Its disabled by default. Iirc it was either busted or really resource intensive

Issue #949, opened 2025-02-11
Apparatus Vivi system


Saturnorsomthing:
* Aka mood n all but packed into one big suggestion

* This dosent go fully into detail, it gives examples and explains how it would work, once properly added overtime stuff would be added to it

* This suggestion has a few parts
>Emotions
>Opinions
>TARDIS Profile (aka the TARDIS itself)
>TARDIS Profiling (aka what the TARDIS thinks of others)
>TARDIS emotional states (aka what the TARDIS does without mood rolls)
>Mood rolls
>TARDIS tantrums

***EMOTIONS/MOOD***

* Emotions are activated by doing a action that increases a % of that said mood, once that mood reaches 100% its activated making it so any other action thats in the same category (negative, neutral, positive) goes towards that mood rather then being used to increase the % of a diffrent one, this gives the mood "moody points".

* The % goes towards a random negative mood in that category so hitting the console wont increase all of the moods but it'll for example only increase angry.

* Moody points are used for the TARDIS to preform certain actions which that mood unlocks.

* Only 1 mood can be active at a time, mood changes in 2 ways
>Wait a MC day, the mood will turn into a random neutral one.
>Do somthing that gives points to the opposing type of a mood, aka if a negative one is active do somthing positive, which will decrease the % of the negative one while the % of the positive one goes up, once the negative one is below the positive one the mood switches.

-NEGATIVE MOODS

ANGRY/HURT/DISSAPOINTED/DEPRESSED/HATEFUL
* Hit the console with the hammer 

DISSANPOINTED/FRUSTRATED
* Miss flight event/let water seep in

HATEFUL/ANGRY
* Crash the tardis 

FEARFUL/ANXIOUS
* Travel to any dimension aside the overworld/locate a end city/fortress (only applys if you dont have pilot loyalty however still still be affected by otheres unless you have owner loyalty) or if a reject is present inside the interior

UPSET/HURT/DISSAPOINTED
* If a bound key gets destroyed 

JELOUS
* If more then 2-3 pets are present in your interior or you bind a key to someone else's tardis

DEPRESSED
* Lack of anyone being in the tardis interior for 2 MC days,massive boost if pilot/owner dies in the interior

-NEUTRAL MOODS

BORED/IMPASSIVE/LONELY
* Lack of anyone being in the tardis interior for 1 MC day

TIERD/MEDIOCRE
* Travel at night time with the tardis (mc has a sun detector or whatever so just copy that code ig)

CURIOUS
[...]

Issue #673, opened 2024-12-08
Ghost monument rework


Saturnorsomthing:
* Occures when
>No one in interior while in flight
>Engine phasing (if you cant stop engine phasing it'll either randomly land somewhere or with a 75% ish chance ghost monument)

* Once ghost monument occures the TARDIS will "split",anyone who has a linked key,sonic or has pilot/owner loyalty will have the TARDIS ghost monument in their area.

* It'll appear for a few pulses before vanishing again.

* If you get close enough you can use a key to stabalize it forcing it to land or use a sonic from a further distance to stabalize it and force landing.

* Once stabalized all the othere "duplicates" vanish.

duzos:
I like this, I'll do it.

PR #1942, opened 2025-10-28
feat: home pos

Peter1169:
## About the PR

- The TARDIS now remembers its grown-spot location as its "home".
- It can be changed by right-clicking with a coral fragment on the telepathic circuits.

## Why / Balance

- This PR only implements the home pos (and reset mechanic) as a tool for other potentially useful gameplay mechanics.
- Like turtles, TARDISes are sentimental; in case of danger or uncertainty, they can return to a familiar place.
- TARDISes dislike the Nether dimension. That is why more loyalty is required to set the home pos in the Nether.
- This could be used in #1381: a TARDIS mats to the home pos for a set amount of time (or waits for an event) before returning to the previous pos if a major danger is present (as in the special “Wild Blue Yonder”). A Dalek raid could be one case.
- If loyalty is low and a mood roll occurs, #949 could trigger the TARDIS to return to the home pos and refuse to leave for some time.
- A Ghost Monument (#673) with no logged player with a key could just return to the home pos.
- As a way to recover lost TARDISes, a TARDIS with no player interaction for a very long time could return to the home pos and power down.
- As a last resort, if a TARDIS exterior ceases to exist for any reason, make it appear at the home pos.
- The home pos is chosen by the player, either where they grow their TARDIS, or later by resetting it via the telepathic circuits. It would be reasonable to implement breaking/removal of any blocks that would prevent the TARDIS from materialising there.

## Technical details

- Added a `home` variable of type `CachedDirectedGlobalPos` in `TravelHandlerBase`, initializing it when a TARDIS is grown, placed, or loaded without a prior home pos.
- Telepathic circuits accept a coral fragment (when not in flight) to reset the home pos to the current location, requiring PILOT loyalty level, or OWNER for the Nether.
- Registered the `/ait home` command to get a TARDIS’s current home pos (coords and dimension) or to set the current pos as the home pos.
- Updated language providers for acceptance and denial messages when changing the TARDIS home pos via telepathic circuits.

## Requirements

## Breaking changes

N/A

**Changelog**
🆑

- Add: TARDIS remembers a "home" pos, either its grown spot or the current location for existing ones. Can be reset to the current location using a coral fragment on the telepathic circuits (requires PILOT loyalty level, or OWNER for the Nether).

### DIFF
--- a/src/main/java/dev/amble/ait/core/commands/HomeCommand.java
+++ b/src/main/java/dev/amble/ait/core/commands/HomeCommand.java
@@ -1,23 +1,33 @@
 package dev.amble.ait.core.commands;
 
-import static net.minecraft.server.command.CommandManager.argument;
-import static net.minecraft.server.command.CommandManager.literal;
-
 import com.mojang.brigadier.Command;
 import com.mojang.brigadier.CommandDispatcher;
+import com.mojang.brigadier.arguments.StringArgumentType;
 import com.mojang.brigadier.context.CommandContext;
 import com.mojang.brigadier.exceptions.CommandSyntaxException;
-
+import dev.amble.ait.AITMod;
+import dev.amble.ait.compat.permissionapi.PermissionAPICompat;
+import dev.amble.ait.core.AITDimensions;
+import dev.amble.ait.core.commands.argument.TardisArgumentType;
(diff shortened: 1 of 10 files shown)


rapbattlegod32:
i really like this

drtheodor (changes requested):

PR #1944, opened 2025-11-02
sonic disguise changer

Mansarde:
## About the PR
This PR adds a way to quickly swap disguises via the sonic screwdriver:
- Set sonic screwdriver to `INTERACTION` mode.
- Put sonic screwdriver in off-hand.
- Using it now on the exterior block of a TARDIS will make it swap to a different disguise (appropriate to the biome, of course).

<ins>NOTE:</ins>
If isomorphic security is active, then it requires the respective loyalty and a linked key.
If it's not active, then it requires at least `PILOT` loyalty (or OP) with the targeted TARDIS.

## Why / Balance
Previously the `/ait debug` command allowed normal players to switch disguises of their TARDISes.
But that command has since been restricted, which means there is no convenient way anymore to switch the disguise of an adaptive exterior besides using the console monitor.

## Technical details
I added a condition check to the Exterior block-entity.

## Media
https://github.com/user-attachments/assets/00040efe-3027-423a-8c77-32ec6aa9472d

## Requirements

**Changelog**
:cl:
- add: Ability to change disguise of adaptive exterior via sonic screwdriver.

### DIFF
--- a/src/main/java/dev/amble/ait/core/blockentities/ExteriorBlockEntity.java
+++ b/src/main/java/dev/amble/ait/core/blockentities/ExteriorBlockEntity.java
@@ -4,6 +4,8 @@ import static dev.amble.ait.core.tardis.handler.InteriorChangingHandler.MAX_PLAS
 
 import java.util.UUID;
 
+import dev.amble.ait.core.tardis.control.impl.SecurityControl;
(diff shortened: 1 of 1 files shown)


duzos (approved):
LGTM

PR #1954, opened 2025-12-14
api!: hums are now translatable

### LINKED ISSUE #1952: make the hums translatable
Currently in the mod the hum gets the name from the `.json` file and not a translation 

example:
`
  "id": "ait:copper"
`

so ingame the hum would just be called `Copper` and can't be translated. it should be smth like `"hum.ait.copper"` so it can be translated. i believe this is down to `buf.writeIdentifier(hum.id());` in the `ClientHumHandler` but i may be wrong

duzos:
closes #1952 
datapacks will now need to define a hum name in their lang files


Loqor (approved):
LGTM

PR #1974, opened 2026-01-24
Datapack Control Animations

### LINKED ISSUE #1953: per control animations for datapack consoles

duzos:
closes #1953


drtheodor (changes requested):

src/main/java/dev/amble/ait/registry/impl/console/variant/ClientConsoleVariantRegistry.java
```diff
@@ -126,7 +162,14 @@ public ConsoleModel model() {
...
 
-                return getSameParent().model();
+	            ClientConsoleVariantSchema parent = getSameParent(val -> val.model() != null);
+	            if (parent != null) {
+		            return parent.model();
+	            }
+
+                AITMod.LOGGER.warn("Datapack console variant {} has no model set, falling back to Hartnell model", variant.id());
+
+	            return HARTNELL.model();
```
drtheodor: won't this re-create the hartnell model each time this gets called in case of a failure?

src/main/java/dev/amble/ait/data/datapack/DatapackConsole.java
```diff
@@ -161,6 +180,18 @@ public Optional<SimpleType> getCustomType() {
...
     }
 
+	public @Nullable ConsoleVariantSchema getSameParent(Predicate<ConsoleVariantSchema> predicate) {
+        if (this.parent() == null) return null;
+
+        Optional<ConsoleVariantSchema> found = ConsoleVariantRegistry.withParent(this.parent()).stream().filter(val -> val != this).filter(predicate).findAny();
+
+        if (found.isPresent()) return found.get();
+
+        AITMod.LOGGER.warn("DatapackConsole {} could not find parent variant matching predicate", this.id());
```
drtheodor: this will get spammed each frame. if this can be logged once during reload that would be great, if not then thats alr

src/main/java/dev/amble/ait/registry/impl/console/variant/ClientConsoleVariantRegistry.java
```diff
@@ -126,7 +162,14 @@ public ConsoleModel model() {
                     return new BedrockConsoleModel(BedrockModelRegistry.getInstance().get(variant.model().get()));
                 }
 
-                return getSameParent().model();
+	            ClientConsoleVariantSchema parent = getSameParent(val -> val.model() != null);
+	            if (parent != null) {
+		            return parent.model();
+	            }
+
+                AITMod.LOGGER.warn("Datapack console variant {} has no model set, falling back to Hartnell model", variant.id());
```
drtheodor: this log will also be spammed, which is alr ig but if it can be logged only once during reloads that would be great

drtheodor (changes requested):

src/main/java/dev/amble/ait/registry/impl/console/variant/ClientConsoleVariantRegistry.java
duzos: yes but models are instanced i believe

drtheodor: i was more thinking like
like yk how in source and garrys mod and wherever itd load that ERROR model as a placeholder
i thought it was smth like that
but ig not
alr then

drtheodor (changes requested):

duzos:
**how to use this**

console datapacks live in `data/<namespace>/console/<name>.json`

give it a `type` with its controls, or a `parent` if its just a variant. each control can take an `animation` now, its a bedrock ref (`model:anim`) pointing at an anim in that models `.animation.json`:

```json
{
  "id": "ait:throttle",
  "width": 0.1875,
  "height": 0.1625,
  "offset": [0.016, 0.75, -0.66],
  "animation": "coral:throttle"
}
```

the anim isnt play once, it scrubs to wherever the control actually is. each control works out its own spot with getTargetProgress, 0 is the start of the anim, 1 is the end:
- handbrake: 1 if engaged else 0
- throttle: speed / max speed, so it sits partway
- door: (left + right) / 2
- anything without an override: 1 on cooldown, 0 otherwise

model + transforms sit on the console:

```json
{
  "id": "ait:coral",
  "type": { "id": "ait:coral", "name": "Coral", "controls": [ ... ] },
  "texture": "ait:textures/blockentities/consoles/coral.png",
  "emission": "ait:textures/blockentities/consoles/coral_emission.png",
  "model": "ait:coral",
  "transformations": { "offset": [0.5, -0.1, -0.5], "scale": [1, 1, 1] }
}
```

transformations takes `offset`, `scale`, `rotation`, all optional.

you dont have to set everything. if model/transforms/animations are missing it grabs them off a sibling with the same parent, falls back to hartnell if theres nothing. so variants can just be texture + emission.

src/main/java/dev/amble/ait/registry/impl/ControlRegistry.java
```diff
@@ -39,6 +37,10 @@ public static Optional<Control> fromId(Identifier id) {
         return Optional.ofNullable(REGISTRY.get(id));
     }
 
+	public static Control fallback() {
+		return REGISTRY.get(0);
+	}
```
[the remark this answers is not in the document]
duzos: resolves the throttle control explicitly now and falls back to the first registered control, and throws with a clear message if the registry is somehow empty instead of returning null

src/main/java/dev/amble/ait/core/tardis/control/ControlTypes.java
```diff
@@ -52,6 +62,13 @@ public String toString() {
     }
 
     public Control getControl() {
+        if (this.control == null) {
+            // print stacktrace
+            Thread.dumpStack();
+            AITMod.LOGGER.error("ControlTypes has null control! Returning fallback control.");
+            return ControlRegistry.fallback();
+        }
```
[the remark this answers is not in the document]
duzos: removed the dumpStack, just logs the error and returns the fallback now

src/main/java/dev/amble/ait/core/entities/ConsoleControlEntity.java
```diff
@@ -136,6 +151,16 @@ public void writeCustomDataToNbt(NbtCompound nbt) {
...
+
+	    // write control type via codec
+	    if (this.controlType == null) {
+		    AITMod.LOGGER.error("Control type is null for control entity at {}", this.getPos());
+		    return;
+	    }
+	    DataResult<NbtElement> result = ControlTypes.CODEC.encodeStart(NbtOps.INSTANCE, this.controlType);
+	    NbtElement nbtResult = result.resultOrPartial(AITMod.LOGGER::error).orElseThrow();
+
+	    nbt.put("controlType", nbtResult);
```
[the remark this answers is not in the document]
duzos: now logs and skips writing the controlType instead of throwing, so malformed data wont kill chunk saving

src/main/java/dev/amble/ait/registry/impl/console/variant/ClientConsoleVariantRegistry.java
```diff
@@ -110,14 +113,47 @@ private ClientConsoleVariantSchema getSameParent() {
...
+
+			        if (schema.parent() == null)
+				        continue;
+
+			        if (schema.parent().id().equals(this.parent().id()) && predicate.test(schema))
+				        return schema;
+		        }
+
+		        return null;
+	        }
```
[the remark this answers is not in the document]
duzos: guarded this.parent() once at the top, returns null if it cant be resolved

src/main/java/dev/amble/ait/core/tardis/control/impl/pos/IncrementControl.java
```diff
@@ -47,4 +48,9 @@ private void messagePlayerIncrement(ServerPlayerEntity player, Tardis tardis) {
     public boolean shouldHaveDelay() {
         return false;
     }
+
+	@Override
+	public float getTargetProgress(Tardis tardis, boolean cooldown, @Nullable ConsoleControlEntity entity) {
+		return IncrementManager.getIncrementPosition(tardis) / (float) IncrementManager.getIncrementSize();
+	}
```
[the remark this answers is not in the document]
duzos: good catch, divides by (size - 1) now with a guard for size <= 1 so the last increment hits 1.0

src/main/java/dev/amble/ait/core/tardis/control/impl/DimensionControl.java
```diff
@@ -82,4 +82,14 @@ private void messagePlayer(ServerPlayerEntity player, ServerWorld world, boolean
...
+
+	@Override
+	public float getTargetProgress(Tardis tardis, boolean cooldown, @Nullable ConsoleControlEntity entity) {
+		// return selected dim / all dims
+		CachedDirectedGlobalPos dest = tardis.travel().destination();
+		int index = WorldUtil.travelWorldIndex(dest.getWorld());
+		int total = WorldUtil.getTravelWorlds().size();
+		if (total == 0) return 0.0f;
+		return (float) index / (float) total;
+	}
```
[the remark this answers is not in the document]
duzos: clamped the index with max(0, ...) and normalize by (total - 1) now, with a guard for total <= 1

src/main/java/dev/amble/ait/client/models/consoles/BedrockConsoleModel.java
```diff
@@ -70,12 +75,37 @@ public void animateBlockEntity(ConsoleBlockEntity console, TravelHandlerBase.Sta
...
+	}
+
+	private void applyControlAnimation(ConsoleControlEntity entity) {
+		if (entity.tardis().isEmpty()) return;
+
+		Control control = entity.getControl();
+		if (control == null) return;
+
+		BedrockAnimation anim = this.animationCache.computeIfAbsent(entity.getControl().id(), k -> findAnimationById(entity));
+		if (anim == null) return;
```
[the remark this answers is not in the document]
duzos: yeah, caches by the animation reference id now instead of the control id

src/main/java/dev/amble/ait/core/entities/ConsoleControlEntity.java
```diff
@@ -290,9 +329,42 @@ public void setControlHeight(float height) {
     }
 
     public Control getControl() {
+	    // On client, look up control from synced ID if local field is null
+	    if (this.control == null && this.getWorld() != null && this.getWorld().isClient()) {
+		    String controlId = this.dataTracker.get(CONTROL_ID);
+		    if (!controlId.isEmpty()) {
+			    this.control = ControlRegistry.fromId(Identifier.tryParse(controlId)).orElse(null);
+		    }
```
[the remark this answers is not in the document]
duzos: guarded it, parses the id first and only looks it up if its non-null now

src/main/java/dev/amble/ait/registry/impl/console/variant/ConsoleVariantRegistry.java
```diff
@@ -65,7 +63,7 @@ public void syncToClient(ServerPlayerEntity player) {
             }
 
             buf.encodeAsJson(DatapackConsole.CODEC, new DatapackConsole(schema.id(), Optional.of(schema.parent().id()),
-                    DatapackExterior.DEFAULT_TEXTURE, DatapackExterior.DEFAULT_TEXTURE, List.of(), new Vector3f(), List.of(), new Vector3f(), Optional.empty(), Vec3d.ZERO, Vec3d.ZERO, null, Optional.empty(), false));
+		            DatapackExterior.DEFAULT_TEXTURE, DatapackExterior.DEFAULT_TEXTURE, List.of(), new Vector3f(), List.of(), new Vector3f(), Optional.empty(), Transformations.DEFAULT, null, Optional.empty(), false));
```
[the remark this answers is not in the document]
duzos: switched to parentId() here, no more parent() resolve

src/main/java/dev/amble/ait/registry/impl/console/variant/ClientConsoleVariantRegistry.java
```diff
@@ -110,14 +113,51 @@ private ClientConsoleVariantSchema getSameParent() {
...
+
+			        if (schema.parent() == null)
+				        continue;
+
+			        if (schema.parent().id().equals(thisParent.id()) && predicate.test(schema))
+				        return schema;
+		        }
+
+		        return null;
+	        }
```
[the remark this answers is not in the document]
duzos: good catch, these client variants get built with parent == id so parent().id() was the variants own id and never matched. compares by parentId() (the console type) now like withSameParent does. also made the texture/emission/model predicates read the raw datapack values instead of calling back into the recursive methods, so fixing the match doesnt reintroduce the stack overflow

src/main/java/dev/amble/ait/data/datapack/DatapackConsole.java
```diff
@@ -57,15 +49,23 @@ public class DatapackConsole extends ConsoleVariantSchema implements TravelAnima
                             .forGetter(DatapackConsole::handlesRotation),
                     MoreCodec.VECTOR3F.optionalFieldOf("handles_translation", new Vector3f()).forGetter(DatapackConsole::handlesTranslation),
                     Identifier.CODEC.optionalFieldOf("model").forGetter(DatapackConsole::model),
-                    Vec3d.CODEC.optionalFieldOf("scale", new Vec3d(1, 1, 1)).forGetter(DatapackConsole::getScale),
-                    Vec3d.CODEC.optionalFieldOf("offset", new Vec3d(0, 0, 0)).forGetter(DatapackConsole::getOffset),
+		            Transformations.CODEC.optionalFieldOf("transformations", Transformations.MISSING).forGetter(DatapackConsole::getTransformations),
                     TravelAnimationMap.CODEC.optionalFieldOf("animations", new TravelAnimationMap())
                             .forGetter(DatapackConsole::getAnimations),
                     SimpleType.CODEC.optionalFieldOf("type").forGetter(DatapackConsole::getCustomType),
                     Codec.BOOL.optionalFieldOf("isDatapack", true).forGetter(DatapackConsole::wasDatapack))
```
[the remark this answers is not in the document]
duzos: added a transition path, fromJson now reads the old top-level scale/offset and builds a Transformations from them when transformations isnt set, so old datapacks keep working. kept it in fromJson instead of the codec so it doesnt eat another codec slot

src/main/java/dev/amble/ait/core/blockentities/ConsoleBlockEntity.java
```diff
@@ -293,6 +296,51 @@ public void killControls() {
...
+	 */
+	public List<ConsoleControlEntity> getControlEntities() {
+		if (this.world == null) return Collections.emptyList();
+
+		// On client, use cached list that refreshes periodically
+		if (this.world.isClient()) {
+			if (cachedClientControls == null || this.age % CLIENT_CACHE_REFRESH_INTERVAL == 0) {
+				cachedClientControls = findControlEntitiesInWorld();
+			}
+			return cachedClientControls;
```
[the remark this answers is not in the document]
duzos: tracks the last refresh age now and only re-runs the query once we're past the interval, so it doesnt re-query every render call on the refresh tick

src/main/java/dev/amble/ait/core/tardis/control/impl/pos/PosControl.java
```diff
@@ -51,7 +49,7 @@ private void messagePlayerDestination(ServerPlayerEntity player, TravelHandler t
 
     @Override
     public boolean shouldHaveDelay() {
-        return false;
+	    return true;
```
drtheodor, line 53: you sure about this

src/main/java/dev/amble/ait/client/models/consoles/BedrockConsoleModel.java
```diff
@@ -70,12 +75,35 @@ public void animateBlockEntity(ConsoleBlockEntity console, TravelHandlerBase.Sta
...
+		ControlTypes type = entity.getControlType().orElse(null);
+		if (type == null) return;
+
+		BedrockAnimationReference ref = type.getAnimation().orElse(null);
+		if (ref == null) return;
+
+		// cache by the animation reference id, not the control id, so a control mapped to different
+		// animations across consoles/variants doesn't reuse the wrong cached animation
+		BedrockAnimation anim = this.animationCache.computeIfAbsent(ref.id(), k -> ref.get().orElse(null));
+		if (anim == null) return;
```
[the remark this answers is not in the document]
duzos: good point, switched to containsKey + put so a missing animation ref gets memoized as null instead of re-resolving every frame

src/main/java/dev/amble/ait/registry/impl/console/variant/ClientConsoleVariantRegistry.java
```diff
@@ -110,14 +113,51 @@ private ClientConsoleVariantSchema getSameParent() {
...
+
             @Override
             public Identifier texture() {
-                return variant.texture();
+	            if (texture == null) {
+		            if (variant.texture().equals(DatapackConsole.EMPTY)) {
+			            ClientConsoleVariantSchema parent = getSameParent(val -> val.parent() instanceof DatapackConsole dp && !dp.texture().equals(DatapackConsole.EMPTY));
```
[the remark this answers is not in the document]
duzos: same deal, intentional. a datapack variant inherits a model from a datapack sibling else falls back to hartnell. not pulling from built-in variants on purpose

src/main/java/dev/amble/ait/registry/impl/console/variant/ClientConsoleVariantRegistry.java
```diff
@@ -126,7 +166,14 @@ public ConsoleModel model() {
                     return new BedrockConsoleModel(BedrockModelRegistry.getInstance().get(variant.model().get()));
                 }
 
-                return getSameParent().model();
+	            ClientConsoleVariantSchema parent = getSameParent(val -> val.parent() instanceof DatapackConsole dp && dp.model().isPresent());
```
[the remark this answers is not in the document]
duzos: same deal, intentional. a datapack variant inherits a model from a datapack sibling else falls back to hartnell. not pulling from built-in variants on purpose

src/main/java/dev/amble/ait/client/renderers/consoles/ConsoleRenderer.java
```diff
@@ -120,7 +117,7 @@ private void renderConsole(Profiler profiler, ClientTardis tardis, T entity, Mat
         profiler.swap("render");
         model.renderWithAnimations(tardis, entity, model.getPart(),
                 matrices, vertexConsumers.getBuffer(variant.equals(ClientConsoleVariantRegistry.COPPER) ? RenderLayer.getEntityTranslucent(variant.texture()) :
-                        RenderLayer.getEntityCutout(variant.texture())), light, overlay,
+				        RenderLayer.getEntityTranslucentCull(variant.texture())), light, overlay,
                 1, 1, 1, 1, tickDelta);
```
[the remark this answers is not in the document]
duzos: intentional, the bedrock console models are authored for translucent rendering

src/main/java/dev/amble/ait/registry/impl/console/variant/ClientConsoleVariantRegistry.java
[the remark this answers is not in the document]
duzos: (correction: my reply above about model landed on the wrong thread) for texture specifically - same intentional limitation, it pulls from a datapack sibling else keeps the variants own texture. built-in inheritance left out to stay recursion-safe

src/main/java/dev/amble/ait/client/models/consoles/BedrockConsoleModel.java
```diff
@@ -71,12 +76,42 @@ public void animateBlockEntity(ConsoleBlockEntity console, TravelHandlerBase.Sta
             throw new IllegalStateException("DatapackConsole " + console.getVariant().id() + " has no animations defined.");
         }
 
-        BedrockAnimation anim = map.getAnimation(state);
-
-        if (anim == null) return;
 
         this.getPart().traverse().forEach(ModelPart::resetTransform);
```
drtheodor: why doesnt this use the funny cached animation reset

drtheodor (changes requested):

src/main/java/dev/amble/ait/core/blockentities/ConsoleBlockEntity.java
```diff
@@ -60,6 +60,10 @@ public class ConsoleBlockEntity extends AbstractConsoleBlockEntity implements Bl
 
     private boolean needsControls = true;
 
+	private static final int CLIENT_CACHE_REFRESH_INTERVAL = 2400; // Lazy refresh
+	// Client-side cache for control entities
+	private List<ConsoleControlEntity> cachedClientControls = null;
```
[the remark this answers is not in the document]
duzos: fair, that cache was jank. gone, along with the box query it was caching.

controls add themselves to the console's list on their own tick now, so client and server read the same list. no scan per console, no 2400 tick staleness window.

entities stay in the render path though. the animation itself is on the model like you said, but the per control animation state and the cooldown flag both live on the entity and only reach the client as tracked data. the console has no synced per control cooldown, so pulling the entities out means building a new sync channel for it first. only sonicport even reads the entity param on getTargetProgress, and only to get back to the console.

the animation lookup memoization is the same deal as the bone cache in #2186, misses cached too so a missing ref doesn't get re-resolved every frame

src/main/java/dev/amble/ait/client/models/consoles/BedrockConsoleModel.java
duzos: no reason, just missed it. walks a cached flat array now, same as SimpleConsoleModel#parts from #2186.

didn't need the root guard that one carries though, the bedrock root is assigned once in the ctor and final, so there's nothing to invalidate against. a datapack resync builds a whole new model anyway

src/main/java/dev/amble/ait/core/tardis/control/impl/pos/PosControl.java
duzos: yeah, deliberate. the default getTargetProgress is `cooldown ? 1 : 0`, so a control with no delay has nothing to animate off and the xyz levers just sit there. flipping it is what makes them move at all.

cost is a 5 tick gate on coordinate nudges, which i'll take over them being static. the proper fix is a separate "recently used" signal so the animation doesn't have to borrow the cooldown, but that's a new synced field and not this PR

src/main/java/dev/amble/ait/registry/impl/ControlRegistry.java
```diff
@@ -39,6 +37,15 @@ public static Optional<Control> fromId(Identifier id) {
         return Optional.ofNullable(REGISTRY.get(id));
     }
 
+	public static Control fallback() {
+		Control fallback = fromId(AITMod.id("throttle")).orElse(REGISTRY.get(0));
```
drtheodor: static final the default (fallback) id (that is, throttle)

src/main/java/dev/amble/ait/core/tardis/control/ControlTypes.java
```diff
@@ -52,6 +62,11 @@ public String toString() {
     }
 
     public Control getControl() {
+        if (this.control == null) {
+            AITMod.LOGGER.error("ControlTypes has null control! Returning fallback control.");
```
drtheodor, line 66: Will this spam the log (does it get called in rendering)?

drtheodor (changes requested):

src/main/java/dev/amble/ait/registry/impl/ControlRegistry.java
duzos: done, FALLBACK_ID static final now.

also caught while in there: the id resolve used `.orElse(ControlRegistry.fallback())`, and orElse evaluates eagerly, so every ControlTypes built from an id was resolving the fallback alongside the control it actually wanted. orElseGet now.

left it private since nothing else needs it, say the word if you want it exposed

src/main/java/dev/amble/ait/core/tardis/control/ControlTypes.java
duzos: no, it's not in the render path. that's ControlTypes#getControl, and the only callers are spawnControls (server, once per console), readCustomDataFromNbt, setControlData and the codec getter.

the renderer calls ConsoleControlEntity#getControl instead, which is a different method that lazily resolves off the synced id and doesn't log. so worst case this fires once per control per load, not per frame

src/main/java/dev/amble/ait/registry/impl/console/variant/ConsoleVariantRegistry.java
```diff
@@ -117,7 +118,7 @@ public static List<ConsoleVariantSchema> withParent(ConsoleTypeSchema parent) {
                 continue;
             }
 
-            if (schema.parent().equals(parent))
+	        if (schema.parent() != null && schema.parent().equals(parent))
```
drtheodor: if `parent` is guaranteed to be non-null (and im not sure it is) then we can do yoda `parent.equals(schema.parent())`

src/main/java/dev/amble/ait/core/entities/ConsoleControlEntity.java
```diff
@@ -82,9 +88,15 @@ public class ConsoleControlEntity extends LinkableDummyEntity {
...
     private static final TrackedData<String> CONTROL_ID = DataTracker.registerData(ConsoleControlEntity.class,
             TrackedDataHandlerRegistry.STRING);
-    private Control control;
+    private static final TrackedData<String> ANIMATION_ID = DataTracker.registerData(ConsoleControlEntity.class,
+            TrackedDataHandlerRegistry.STRING);
     public static final float MAX_DURABILITY = 1.0f;
 
+    private Control control;
+    private ControlTypes controlType;
+    private ConsoleBlockEntity linkedConsole;
```
drtheodor: what if the console gets nuked without setRemoved getting called?

src/main/java/dev/amble/ait/core/entities/ConsoleControlEntity.java
```diff
@@ -105,6 +117,11 @@ public void onRemoved() {
             return;
         }
 
+        if (this.linkedConsole != null) {
+            this.linkedConsole.controlEntities.remove(this);
```
drtheodor: praying onRemoved gets called and the console block pos gets synced properly

src/main/java/dev/amble/ait/core/entities/ConsoleControlEntity.java
drtheodor: on a slightly unreltaed note, why do we keep the console block pos in a data tracker? it wouldnt change in real time, cant we keep it in the nbt (same for width and height ig)?

src/main/java/dev/amble/ait/client/models/consoles/BedrockConsoleModel.java
```diff
@@ -71,12 +95,43 @@ public void animateBlockEntity(ConsoleBlockEntity console, TravelHandlerBase.Sta
...
 
-        this.getPart().traverse().forEach(ModelPart::resetTransform);
+		BedrockAnimation anim = map.getAnimation(state);
+		if (anim == null) return;
 
-        anim.apply(this.getPart(), console.ANIM_STATE, console.getAge(), 1F, null);
+		anim.apply(this.getPart(), console.ANIM_STATE, console.getAge(), 1F, null);
     }
+
+	private void applyControlAnimation(ConsoleControlEntity entity) {
```
drtheodor: so the question i have is why do we need a reference to an actual entity to do this, given that we just need the Control instance (and afaik those are static)

can't we just get the console type and then get all the controls it has?

src/main/java/dev/amble/ait/client/models/consoles/BedrockConsoleModel.java
```diff
@@ -71,12 +95,43 @@ public void animateBlockEntity(ConsoleBlockEntity console, TravelHandlerBase.Sta
...
+
+		BedrockAnimationReference ref = type.getAnimation().orElse(null);
+		if (ref == null) return;
+
+		// cache by the animation reference id, not the control id, so a control mapped to different
+		// animations across consoles/variants doesn't reuse the wrong cached animation
+		// memoize misses too (computeIfAbsent won't store null) so a missing animation ref isn't re-resolved every frame
+		BedrockAnimation anim;
+		if (this.animationCache.containsKey(ref.id())) {
+			anim = this.animationCache.get(ref.id());
```
drtheodor: micro-optimization: get without a contains check
as for the cache map you can use an optional for the values since you need to indicate a missing animation
alternatively, if theres some default static constant empty animation you can use that instead of wrapping stuff around optionals

drtheodor (changes requested):

src/main/java/dev/amble/ait/client/models/consoles/BedrockConsoleModel.java
duzos: done, the model reads the type schema now and never touches a control entity. walks `console.controlTypes()` by index, animation progress lives per slot on the console, and the console implements EffectProvider so fx still have a source.

by index rather than by Control because copper lists handbrake 3x and throttle 2x, and `apply` calls `setAnimationLength` on the state it's given, so sharing one state between them would have had the two animations fighting over it.

the one thing i couldn't move is the cooldown. it's what the default `getTargetProgress` animates off (`cooldown ? 1 : 0`) and it only reaches the client as tracked data on the entity. console side would mean syncing it, and `toUpdatePacket` is a full NBT resend, so every press would resend the console twice. i know the generator sends its own SYNC_TYPE/SYNC_VARIANT payloads so a bespoke packet is buildable, it's just more machinery than a tracked boolean that already exists. so the entity keeps the flag and pushes it to the console while it's set and once on release, and the model only ever reads console state.

also dropped `getControlType`, `ANIMATION_ID` and `implements AnimatedEntity` off the entity since nothing needed them after that

src/main/java/dev/amble/ait/core/entities/ConsoleControlEntity.java
duzos: both gone, along with the thing they were about. no block entity reference is held across ticks any more and nothing depends on `onRemoved` firing: the entity pushes to the console rather than being read from it, so a console that disappears without warning just stops receiving pushes, and the console expires an unrenewed cooldown on its own.

while in there, `onRemoved`'s `getConsoleBlockPos() == null` guard was dead anyway, CONSOLE_BLOCK_POS defaults to BlockPos.ORIGIN so it's never null, which meant `super.onRemoved()` never ran on the live path. fixed

src/main/java/dev/amble/ait/core/entities/ConsoleControlEntity.java
duzos: see the thread above, the praying is gone. nothing depends on onRemoved now

src/main/java/dev/amble/ait/core/entities/ConsoleControlEntity.java
duzos: nbt doesn't reach the client at all here, the entity uses a plain EntityType.Builder so the client only ever sees the spawn packet and the trackers. so it'd have to be a fabric custom spawn packet, not nbt.

agreed on the principle though. of the 12 tracked fields 6 are spawn time constants (console pos, width, height, offset, control id, and the index i just added) and 5 actually change. worth doing as its own change, it's entity networking and nothing to do with console rendering.

this refactor cuts into it a bit already: CONTROL_ID and ANIMATION_ID are both gone, replaced by one int index

src/main/java/dev/amble/ait/client/models/consoles/BedrockConsoleModel.java
duzos: done, single get with Optional values so a miss is a cached empty rather than an absent key. no static empty animation to point at, so Optional it is.

that method got rewritten anyway since the model stopped taking entities, it's `animation(ref)` now

src/main/java/dev/amble/ait/registry/impl/console/variant/ConsoleVariantRegistry.java
duzos: dropped the redundant `schema.parent() != null` instead, the block above already continues on a null parent so it was dead.

left the direction as is rather than going yoda: `withParent`'s own argument can be null (ConsoleBlockEntity passes `current.parent()` unguarded), and `parent.equals(...)` would NPE on that where `schema.parent().equals(parent)` just returns false

PR #2001, opened 2026-03-23
Whitelist alternatives in the config

Peter1169:
## About the PR
Adds alternative whitelists for the four existing dimension blacklists:
- Environment Projector
- TARDIS travel
- Rift spawn
- Rift drop

If a whitelist has entries, it overrides its matching blacklist.

## Why / Balance
This gives modpack/server owners a cleaner allow-only configuration for dimensions. Large modpacks no longer have to manually enter 47 dimension entries in the blacklist and can just use the whitelist instead.
Balance is unchanged unless these new config entries are used, since all whitelists default to empty.

## Technical details
Added four new server config lists and their config screen translations.
Updated world cache generation so whitelist entries take priority over blacklists, while ignoring empty/invalid entries safely.
Also added a few guards so empty resolved lists do not cause projector, travel control, or rift-related runtime issues.

## Media
N/A

## Requirements

## Breaking changes
N/A

**Changelog**
:cl:
- tweak: Alternative dimension whitelists in the config for the Environment Projector, TARDIS travel, rift spawn, and rift drop; whitelists override their matching blacklists when populated.


drtheodor (commented):
LGTM, tested?

Peter1169:
yes

duzos (approved):
LGTM

PR #2003, opened 2026-04-25
feat: scrolling astral map screen

nanowuu:
## About the PR

- Added a scrollable astral map list which can be searched by name or by mod ID (with @)
- Added a separate translation key for biomes and structures not being found

## Why / Balance
Better for usability

## Technical details
- Astral Map screen completely changed to use a scrollable and searchable list widget, no more IdentifierSwitcher
- Removed biome synchronisation (that data is already available on the client, maybe because of the F3 screen)
- Added CallbackCheckboxWidget which allows for a callback when the checkbox is checked

## Media

image
image

## Requirements

- [X ] I have read and am following the [Pull Request and Changelog Guidelines](https://amblelabs.github.io/ait-wiki/guidelines).
- [ X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

No more IdentifierSwitcher or AstralMapBlock.biomeIds but they weren't used by anything else anyway

**Changelog**

:cl:
- Changed astral map GUI to be a scrollable and searchable list: prefix searches with @ to search by mod ID


rapbattlegod32:
ill test now

rapbattlegod32:
my intellij is bugging out i cant test 
@Addi3 you're going to have to review instead

addi3phone:
> my intellij is bugging out i cant test @Addi3 you're going to have to review instead

Will be in like 1hr or smth cus I'm not on my pc rn

Addi3 (approved):
LGTM. Tested in game, locates structures and biomes fine, no noticeable issues with the GUI either and scales well with game window

src/main/java/dev/amble/ait/client/screens/AstralMapScreen.java
```diff
@@ -103,123 +60,257 @@ public boolean shouldPause() {
...
-                this.textRenderer.getWidth(Text.translatable("screen.ait.astral_map.biomes.button")),
-                10,
-                Text.translatable("screen.ait.astral_map.biomes.button"),
-                button -> biomeFinderButton(),
-                this.textRenderer
-        ));
+        this.searchBox = new TextFieldWidget(this.client.textRenderer, this.left + 12,
+                this.top + 13, this.bgWidth - 26, 15, SEARCH_TEXT);
+        this.searchBox.setPlaceholder(SEARCH_TEXT);
+        this.searchBox.setText("");
```
drtheodor: Isn't it empty by default?

src/main/java/dev/amble/ait/client/screens/AstralMapScreen.java
```diff
@@ -103,123 +60,257 @@ public boolean shouldPause() {
...
-            context.drawText(this.textRenderer, currentText, (int) (left + (bgWidth * 0.5f)) - this.textRenderer.getWidth(currentText) / 2,
-                    (int) (top + (bgHeight * 0.5)), 0xffffff, true);
+    /**
+     * Widget to display a searchable list of biomes and structures to locate
+     */
+     class AstralMapListWidget extends AlwaysSelectedEntryListWidget<AstralMapListWidget.Entry> {
+
+        private DisplayedCategories shownCategory = DisplayedCategories.BOTH;
+        private final List<Entry> entries = new ArrayList<>();
+        private final Map<String, String> mods = new HashMap<>();
```
drtheodor, line 156: Wouldn't it be better if this was static? This way it doesn't have to re-think what the mods' names are. It could be lazily initialized to avoid querying mods for no reason.

src/main/java/dev/amble/ait/client/screens/AstralMapScreen.java
```diff
@@ -103,123 +60,257 @@ public boolean shouldPause() {
...
 
-    private void drawBackground(DrawContext context) {
-        context.drawTexture(TEXTURE, left, top, 0, 0, bgWidth, bgHeight);
-    }
+        public String getModName(String modId) {
+            if (mods.containsKey(modId)) return mods.get(modId);
+            String modName = FabricLoader.getInstance().getModContainer(modId)
+                    .map(mod -> mod.getMetadata().getName()).orElse(modId);
+            mods.put(modId, modName);
+            return modName;
```
drtheodor: All of this could use `Map#computeIfAbsent`:

```suggestion
            return mods.computeIfAbsent(modId, id -> FabricLoader.getInstance().getModContainer(id)
                    .map(mod -> mod.getMetadata().getName()).orElse(id));
```

src/main/java/dev/amble/ait/client/screens/AstralMapScreen.java
```diff
@@ -103,123 +60,257 @@ public boolean shouldPause() {
...
+            if (shownCategory != DisplayedCategories.STRUCTURES) {
+                for (Identifier id : client.world.getRegistryManager().get(RegistryKeys.BIOME).getIds()) {
+                    this.entries.add(new Entry(id, true));
+                }
+            }
+            if (shownCategory != DisplayedCategories.BIOMES) {
+                for (Identifier id : AstralMapBlock.structureIds) {
+                    this.entries.add(new Entry(id, false));
+                }
+            }
```
drtheodor: This is rather confusing and will become more complicated if more filters are added.
I suggest the following:
```suggestion
            boolean both = DisplayedCategories.BOTH;
            if (shownCategory == DisplayedCategories.BIOMES || both) {
                for (Identifier id : client.world.getRegistryManager().get(RegistryKeys.BIOME).getIds()) {
                    this.entries.add(new Entry(id, true));
                }
            }
            if (shownCategory == DisplayedCategories.STRUCTURES || both) {
                for (Identifier id : AstralMapBlock.structureIds) {
                    this.entries.add(new Entry(id, false));
                }
            }
```

Alternatively, the enum could be replaced with a bitset (be it a normal `int` or [`BitSet`](https://docs.oracle.com/javase/8/docs/api/java/util/BitSet.html), effectively allowing to combine values).

Alternatively no.2, you could add an abstract method to the enum that would supply the entries, something like this:

```java
public enum DisplayedCategories {
    STRUCTURES {
        @Override
        public void getEntries(Consumer<AstralMapListWidget.Entry> consumer, MinecraftClient client) {
            for (Identifier id : AstralMapBlock.structureIds) {
                consumer.accept(new Entry(id, false));
            }
        }
    },
    BIOMES {
        @Override
        public void getEntries(Consumer<AstralMapListWidget.Entry> consumer, MinecraftClient client) {
            for (Identifier id : client.world.getRegistryManager().get(RegistryKeys.BIOME).getIds()) {
                consumer.accept(new Entry(id, true));
            }
        }
    },
    BOTH {
        @Override
        public void getEntries(Consumer<AstralMapListWidget.Entry> consumer, MinecraftClient client) {
            for (DisplayedCategories category : DisplayedCategories.VALUES) {
                category.getEntries(consumer, client);
            }
        }
    };

    private static final DisplayedCategories VALUES = values();

    public abstract void getEntries(Consumer<AstralMapListWidget.Entry> consumer, MinecraftClient client);
}

// ...

// in #refreshEntries:
this.shownCategory.getEntries(this.entries::add, client);
```

src/main/java/dev/amble/ait/client/screens/AstralMapScreen.java
```diff
@@ -103,123 +60,257 @@ public boolean shouldPause() {
...
+        }
+
+        public void setCurrentSearch(String currentSearch) {
+            this.currentSearch = currentSearch;
+            this.refreshEntries();
+            String[] splitSearch = currentSearch.split(" ");
+            for (String substring : splitSearch) {
+                if (substring.isEmpty()) break;
+                // Allow users to search by mod by prefixing their search with "@"
+                if (substring.charAt(0) == '@') {
```
drtheodor: This check doesn't really make sense here, it should be taken outside of the loop, since entries can't have 2 different namespaces.

src/main/java/dev/amble/ait/client/screens/AstralMapScreen.java
```diff
@@ -103,123 +60,257 @@ public boolean shouldPause() {
...
+                } else {
+                    this.entries.removeIf(entry -> !entry.text.toString().toLowerCase(Locale.ROOT)
+                            .contains(substring.toLowerCase(Locale.ROOT)));
+                }
+            }
+            this.replaceEntries(this.entries);
+            if (!this.children().isEmpty()) {
+                this.setFocused(this.getFirst());
+            }
+            if (!currentSearch.isEmpty()) {
```
drtheodor: This check seems kind of pointless..?

src/main/java/dev/amble/ait/client/screens/AstralMapScreen.java
drtheodor: Oh I've just realised that it does make sense since we'd want it to be in any order

src/main/java/dev/amble/ait/client/screens/AstralMapScreen.java
```diff
@@ -103,123 +60,257 @@ public boolean shouldPause() {
...
+            this.currentSearch = currentSearch;
+            this.refreshEntries();
+            String[] splitSearch = currentSearch.split(" ");
+            for (String substring : splitSearch) {
+                if (substring.isEmpty()) break;
+                // Allow users to search by mod by prefixing their search with "@"
+                if (substring.charAt(0) == '@') {
+                    if (substring.length() < 2) continue;
+                    this.entries.removeIf(entry -> !entry.identifier.getNamespace()
+                            .contains(substring.substring(1).strip().toLowerCase(Locale.ROOT)));
```
drtheodor:
```suggestion
                    // a "@" and at least one more character
                    if (substring.length() < 2) continue;
                    
                    String modId = substring.substring(1);
                    this.entries.removeIf(entry -> !entry.identifier.getNamespace()
                            .contains(modId));
```

drtheodor: We can skip `String#strip` here since we split on spaces.

src/main/java/dev/amble/ait/client/screens/AstralMapScreen.java
```diff
@@ -103,123 +60,257 @@ public boolean shouldPause() {
...
+            }
+            this.entries.sort((e1, e2) -> e1.text.getString().compareToIgnoreCase(e2.text.getString()));
+        }
+
+        public void setCurrentSearch(String currentSearch) {
+            this.currentSearch = currentSearch;
+            this.refreshEntries();
+            String[] splitSearch = currentSearch.split(" ");
+            for (String substring : splitSearch) {
+                if (substring.isEmpty()) break;
```
drtheodor: Not sure this would ever get hit..?

src/main/java/dev/amble/ait/client/screens/AstralMapScreen.java
```diff
@@ -103,123 +60,257 @@ public boolean shouldPause() {
...
+                // Allow users to search by mod by prefixing their search with "@"
```
drtheodor, line 201:
```suggestion
                substring = substring.toLowerCase(Locale.ROOT);
                
                // Allow users to search by mod by prefixing their search with "@"
```

src/main/java/dev/amble/ait/client/screens/AstralMapScreen.java
```diff
@@ -103,123 +60,257 @@ public boolean shouldPause() {
...
+                if (substring.charAt(0) == '@') {
+                    if (substring.length() < 2) continue;
+                    this.entries.removeIf(entry -> !entry.identifier.getNamespace()
+                            .contains(substring.substring(1).strip().toLowerCase(Locale.ROOT)));
+                } else {
+                    this.entries.removeIf(entry -> !entry.text.toString().toLowerCase(Locale.ROOT)
+                            .contains(substring.toLowerCase(Locale.ROOT)));
```
drtheodor:
```suggestion
                            .contains(substring));
```

src/main/java/dev/amble/ait/client/screens/AstralMapScreen.java
```diff
@@ -103,123 +60,257 @@ public boolean shouldPause() {
...
+                this.modName = getModName(identifier.getNamespace());
+            }
+
+            public static String identifierToName(Identifier id) {
+                try {
+                    return WorldUtil.fakeTranslate(id.getPath());
+                } catch (Exception e) {
+                    return id.toString();
+                }
+            }
```
drtheodor, line 281: `WorldUtil#fakeTranslate` shouldn't throw exceptions. If it does, please wrap it in a try-catch.

src/main/java/dev/amble/ait/client/screens/AstralMapScreen.java
```diff
@@ -103,123 +60,257 @@ public boolean shouldPause() {
...
+            public Entry(Identifier identifier, boolean isBiome) {
+                this.identifier = identifier;
+                // Only biomes have actual translation keys and some modded ones might not
+                if (isBiome) {
+                    String translationKey = identifier.toTranslationKey("biome");
+                    if (Language.getInstance().hasTranslation(translationKey)) {
+                        this.text = Text.translatable(translationKey);
+                    } else {
+                        this.text = Text.literal(identifierToName(identifier));
+                    }
```
drtheodor:
```suggestion
                    this.text = Text.translatableWithFallback(translationKey, identifierToName(identifier));
```

src/main/java/dev/amble/ait/client/screens/AstralMapScreen.java
```diff
@@ -103,123 +60,257 @@ public boolean shouldPause() {
...
+        /**
+         * Entry representing a biome or structure
+         */
+        class Entry extends AlwaysSelectedEntryListWidget.Entry<Entry> {
+
+            private final Identifier identifier;
+            private final Text text;
+            private final String modName;
+
+            public Entry(Identifier identifier, boolean isBiome) {
```
drtheodor: Having a hardcoded `isBiome` is kinda bad, since there could be *other* types added in future that could also have a translation.

In this scenario, I think the enum abstract method option I've mentioned previously would work the best, since then you could pass the `Text` directly to the `Entry`, instead of the `Entry` coming up with it.

src/main/java/dev/amble/ait/core/blocks/AstralMapBlock.java
```diff
@@ -76,12 +75,12 @@ public class AstralMapBlock extends BlockWithEntity implements BlockEntityProvid
...
-                boolean isBiome = buf.readBoolean();
+                boolean isStructure = buf.readBoolean();
 
-                if (isBiome) {
-                    handleBiomeRequest(player, target);
-                } else {
+                if (isStructure) {
                     handleStructureRequest(player, target);
+                } else {
+                    handleBiomeRequest(player, target);
```
drtheodor: Should be an enum, ideally.

src/main/java/dev/amble/ait/core/blocks/AstralMapBlock.java
```diff
@@ -202,40 +198,13 @@ private static void sendStructures(ServerWorld world, ServerPlayerEntity target)
...
-        });
-
-        ClientPlayNetworking.registerGlobalReceiver(SYNC_BIOMES, (client, handler, buf, responseSender) -> {
-            List<Identifier> ids = buf.readList(PacketByteBuf::readIdentifier);
-            client.execute(() -> {
-                AstralMapBlock.biomeIds = ids;
-                if (client.currentScreen instanceof AstralMapScreen screen) {
-                    screen.reloadData();
-                }
+                client.setScreen(new AstralMapScreen());
```
drtheodor, line 206: Why is it setting the screen in `SYNC_STRUCTURES`..?

drtheodor: Oh, I see.
Probably should rename it to `SYNC_ASTRAL_MAP_DATA` or `SYNC_LEVEL_REGISTRIES_IDS`  or `OPEN_ASTRAL_MAP` (probably the best option right now) or something like that, in case we add anything else

drtheodor (changes requested):

duzos:
LARP final boss

src/main/java/dev/amble/ait/client/screens/AstralMapScreen.java
nanowuu: I just left this bit as is from @duzos initial astral map commit

src/main/java/dev/amble/ait/client/screens/AstralMapScreen.java
nanowuu: It does, actually, if you prefix your search with a lot of spaces: empty strings that aren't trailing don't get removed with split(). This should be continue, not break, anyway.

src/main/java/dev/amble/ait/client/screens/AstralMapScreen.java
nanowuu: Generally I would agree with you but what other types? I feel like the enum abstract method complicates things a bit too much for a hypothetical

nanowuu: Unless you have something in mind?

drtheodor: It doesn't complicate stuff that much, plus it allows to add other categories more easily. For example, we could add the home location as a category. Or, perhaps, the waypoints stored in waypoint banks.

Obviously, out of scope for this PR, but doing the enum abstract method and passing `Text`s directly would make it implementable without many changes.

Addi3: > It doesn't complicate stuff that much, plus it allows to add other categories more easily. For example, we could add the home location as a category. Or, perhaps, the waypoints stored in waypoint banks.


never happening btw :al_clueless:

src/main/java/dev/amble/ait/client/screens/AstralMapScreen.java
nanowuu: I would prefer keeping data like this associated with the screen instance - there are likely only going to be a few mods that create new structures or biomes, and players aren't going to be opening the astral map that much anyway. I don't think this needs to be constantly in memory.

src/main/java/dev/amble/ait/client/screens/AstralMapScreen.java
nanowuu: I went with the bitset method

drtheodor (approved):
LGTM

PR #2004, opened 2026-05-01
fix: refactor portal world handling and remove postInit

drtheodor:
## About the PR

This PR fixes a weird bug caused by a fuck-up in TARDIS component sorting, which would cause a crash.

## Why / Balance

This is a serious crash that would happen with players playing with IP. Technically, it could cause either the doors to not work OR a crash would happen. Either way, yikes.

## Technical details
This bug caused `PortalsHandler#postLoad` to be called _before_ `TravelHandlerBase#postLoad`, meaning the levels were `null` during initialization. `EntityRef` requires a world instance, otherwise it'd crash with a NPE.

Both of those were fixed, `EntityRef` now checks if the world is null and returns null beforehand if that's the case. `PortalsHandler` sets the world if it isnt set lazily.

## Requirements

## Breaking changes

**Changelog**


drtheodor:
wow great "lgtm"s everyone gonna merge right away

drtheodor:
(was tested by @max7603 so should be alright)

PR #2005, opened 2026-05-02
fix: don't crash if already crashing during remat

drtheodor:
## About the PR

This PR fixes exterior being placed twice as a result of crashing.

## Why / Balance

Exterior duplication is bad.

## Technical details

Apparently, `#forceRemat` may be called twice. And as it turns out, `#crash` also calls `#forceRemat`. And `#forceRemat` calls `#crash` if remat event fails.

A scenario in which `#forceRemat` was being called twice caused the `SafePosSearch#wrapSafe` method be called twice.

Fortunately, we already keep track whether we're calculating the safe pos search (via the `waiting` field).

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: no more exterior duplication during crashing


drtheodor (commented):
was tested too so

PR #2006, opened 2026-05-02
fix: simplify sucking condition & suck harder

drtheodor:
## About the PR
This PR fixes a problem of the door not sucking hard enough.

## Why / Balance
Because it sucks.

## Technical details
Simplified the suck conditions and change it to use the proper suck offset. Apparently, the direction was reversed, so the sign of the offset has to be positive.

Also, made it use `#toCenterPos`, instead of creating `Vec3d` from the `BlockPos`' coordinates, since that would create a vector on the corner of the block which is not the intended behavior.

## Requirements

**Changelog**

:cl:
- fix: suck harder & proper


Addi3 (approved):
LGTM - it can now suck again

drtheodor:
and suck it will

PR #2014, opened 2026-05-18
fix: controls no longer require the subsystem they depend on to be enabled if they're a part of the flight sequence

drtheodor:
## About the PR

This PR fixes a bug where controls that are a part of the flight sequence would not be interactable.

## Why / Balance

Otherwise, some events would be a guaranteed failure.

## Technical details

The subsystem requirement check was done in `#canRun`, while the sequence handling was done in `#runServer`, which was never called, since `#canRun` failed.

The fix adds a check in `#canRun` to ignore the subsystem dependency if the control is a part of the sequence. This should not break existing behavior, as `#runServer` exits early if the control is a part of the sequence.

## Requirements

## Breaking changes

None

**Changelog**

:cl:
- fix: controls no longer require the subsystem they depend on to be enabled if they're a part of the flight sequence

### DIFF
--- a/src/main/java/dev/amble/ait/core/tardis/control/Control.java
+++ b/src/main/java/dev/amble/ait/core/tardis/control/Control.java
@@ -140,7 +140,7 @@ public class Control implements Identifiable {
 
         SubSystem.IdLike dependent = this.requiredSubSystem();
 
-        if (dependent != null) {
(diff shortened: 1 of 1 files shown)


rapbattlegod32 (approved):
LGTM

PR #2064, opened 2026-06-08
Add Simplified Chinese and Traditional Chinese

smallmoss233:
This branch adds Simplified Chinese and Traditional Chinese translations, along with corresponding world warnings.


vmbbi (approved):
LGTM, its better than nothing

PR #2069, opened 2026-06-08
fix: coral fan

Addi3:
## About the PR
I changed the wall "block" to be a blockstate rather than its own block because this is just a decorative block and does have any updates like the normal coral fans, i also fixed some inconsistencies with the corals "required tool"

## Why / Balance
Since the coral fan block was released, it had issues with not placing on walls correctly, having the wrong collision shape, sometimes not updating / breaking when the  block below was broke

## Technical details

- removed `TardisCoralParentBlock` and `TardisCoralWallBlock` class and `tardis_coral_wall_fan.json` because im not using them anymore, i also fully removed the `TARDIS_CORAL_WALL_FAN` as its now a blockstate
- fixed the inconsistencies with the with the corals "required tool", they are all now `IRON` and not 3 different levels
- removed `provider.registerCoralFanBlock` from `AITDataGenerator` class as its technically not a coral fan block

## Media
image

## Requirements

## Breaking changes
Coral fan wall is now a blockstate rather than its own block. why? because this fan isnt like a normal minecraft coral fan (as in it doesnt change to a "dead" state, it just imitates the placement of a coral fan) :shrug: 

**Changelog**

:cl:
- fix: coral fan placement
- fix: inconsistencies with the with the corals "required tool"


src/main/java/dev/amble/ait/core/blocks/TardisCoralFanBlock.java
```diff
@@ -1,61 +1,120 @@
...
         return state;
     }
-}
+
+    @Override
+    public FluidState getFluidState(BlockState state) {
+        if (state.get(WATERLOGGED)) {
+            return Fluids.WATER.getStill(false);
+        }
+        return super.getFluidState(state);
```
drtheodor:
```suggestion
        return state.get(WATERLOGGED) ? Fluids.WATER.getStill(false) : super.getFluidState(state);
```

src/main/java/dev/amble/ait/core/blocks/TardisCoralFanBlock.java
```diff
@@ -1,61 +1,120 @@
...
+                                                BlockState neighborState, WorldAccess world, BlockPos pos, BlockPos neighborPos) {
         if (state.get(WATERLOGGED)) {
             world.scheduleFluidTick(pos, Fluids.WATER, Fluids.WATER.getTickRate(world));
         }
+
+        if (direction == state.get(FACING).getOpposite()) {
+            if (!state.canPlaceAt(world, pos)) {
+                return Blocks.AIR.getDefaultState();
+            }
+        }
```
drtheodor:
```suggestion
        if (direction == state.get(FACING).getOpposite() && !state.canPlaceAt(world, pos)) {
            return Blocks.AIR.getDefaultState();
        }
```

src/main/java/dev/amble/ait/core/blocks/TardisCoralFanBlock.java
```diff
@@ -1,61 +1,121 @@
...
+        BlockPos attachedPos = pos.offset(face.getOpposite());
+        BlockState attachedState = world.getBlockState(attachedPos);
+
+        if (canPlaceOn(attachedState, face.getOpposite())) {
+            return this.getDefaultState()
+                    .with(FACING, face)
+                    .with(WATERLOGGED, ctx.getWorld().getFluidState(ctx.getBlockPos()).getFluid() == Fluids.WATER);
+        }
 
+        for (Direction direction : Direction.values()) {
```
drtheodor: cache the value of Direction.values in a private static final Direction[] DIRECTIONS field OR iterate through all variants of the FACING property!

drtheodor (changes requested):

src/main/java/dev/amble/ait/core/blocks/TardisCoralFanBlock.java
rapbattlegod32: bump

drtheodor (approved):
LGTM

PR #2077, opened 2026-06-10
Fix: Console Animations

Loqor:
## About the PR
In the first commit, I replaced all calls of `MinecraftClient.getInstance()` with a static final field version as that causes some overhead. In the second, I ended up rewriting the way that consoles do their smooth animations. 

## Why / Balance
Smooth control animations give weight to a more solid feeling experience, as opposed to the previous snappy controls. This PR fixes the previous attempt at this, where the entire console would freak out if another console with a different model was present. This was fixed.

## Technical details
- Replace calls of `MinecraftClient.getInstance()` with a static final field.
- In `SimpleConsoleModel`, add a new `WeakHashMap` that stores the `ConsoleBlockEntity` along with another map, `Map<String, Float>` which stores the "id" of an "animation" with its lerped value, as they typically are just floats (and have been)

## Requirements

## Breaking changes
None

**Changelog**

:cl:
- feat: Smooth control animation across all consoles!

### DIFF
--- a/src/main/java/dev/amble/ait/client/AITModClient.java
+++ b/src/main/java/dev/amble/ait/client/AITModClient.java
@@ -110,6 +110,7 @@ import dev.amble.ait.registry.impl.exterior.ClientExteriorVariantRegistry;
 public class AITModClient implements ClientModInitializer {
 
     public static AITClientConfig CONFIG;
+    public static final MinecraftClient client = MinecraftClient.getInstance();
 
     @Override
     public void onInitializeClient() {
@@ -501,7 +502,6 @@ public class AITModClient implements ClientModInitializer {
     public void exteriorBOTI(WorldRenderContext context) {
         if (skipBuiltInBOTI()) return;
 
-        MinecraftClient client = MinecraftClient.getInstance();
         if (client.player == null || client.world == null) return;
         ClientWorld world = client.world;
         MatrixStack stack = context.matrixStack();
@@ -534,7 +534,6 @@ public class AITModClient implements ClientModInitializer {
     public void doorBOTI(WorldRenderContext context) {
         if (skipBuiltInBOTI()) return;
 
-        MinecraftClient client = MinecraftClient.getInstance();
         if (client.player == null || client.world == null) return;
         ClientWorld world = client.world;
         MatrixStack stack = context.matrixStack();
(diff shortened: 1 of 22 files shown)


src/main/java/dev/amble/ait/client/models/consoles/SimpleConsoleModel.java
```diff
@@ -23,6 +27,28 @@
 @SuppressWarnings("rawtypes")
 public abstract class SimpleConsoleModel extends SinglePartEntityModel implements ConsoleModel {
 
+    protected static final Map<BlockEntity, Map<String, Float>> ANIMATION_CACHE = new WeakHashMap<>();
```
drtheodor: hiss
yikes man
why do you need this map if you store the current rotation in the roll/pitch/yaw?

src/main/java/dev/amble/ait/client/renderers/machines/GenericSubSystemRenderer.java
```diff
@@ -18,6 +18,7 @@
 
 public class GenericSubSystemRenderer<T extends GenericStructureSystemBlockEntity> implements BlockEntityRenderer<T> {
     private final GenericSubSystemModel model;
+    public static final MinecraftClient client = MinecraftClient.getInstance();
```
drtheodor:
```suggestion
    private static final MinecraftClient client = MinecraftClient.getInstance();
```

src/main/java/dev/amble/ait/client/util/FoggyUtils.java
```diff
@@ -21,8 +21,9 @@
 import dev.amble.ait.module.planet.core.space.system.Space;
 
 public class FoggyUtils {
+    public static final MinecraftClient mc = MinecraftClient.getInstance();
```
drtheodor:
```suggestion
    private static final MinecraftClient mc = MinecraftClient.getInstance();
```

src/main/java/dev/amble/ait/client/AITModClient.java
```diff
@@ -110,6 +110,7 @@
 public class AITModClient implements ClientModInitializer {
 
     public static AITClientConfig CONFIG;
+    public static final MinecraftClient client = MinecraftClient.getInstance();
```
drtheodor:
```suggestion
    private final MinecraftClient client = MinecraftClient.getInstance();
```

drtheodor (dismissed):

src/main/java/dev/amble/ait/client/models/consoles/SimpleConsoleModel.java
Loqor: Thats... not how that works. the roll/pitch/yaw stuff is really overly complicated and doesnt work like that. believe me, i tried.

drtheodor (approved):

PR #2090, opened 2026-06-12
Fix: Hypercube Crashing

Loqor:
## About the PR
hypercubes used to crash players on servers. BUT NO MORE I SAY! With these fixes (if you can even call them that) it shouldn't AT LEAST crash players anymore.

## Why / Balance
Crashing bad?

## Technical details
use `ItemStack#removeCustomName()` instead of setting the custom name to null.

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: hypercubes no longer crash players


drtheodor (approved):
LGTM

PR #2080, opened 2026-06-10
perf: console exterior model cache

drtheodor:
## About the PR
This PR fixes a long going performance issue with faulty model caching.

## Why / Balance

Because performance and #2077 

## Technical details

In PR #2077 the code had to be massively overcomplicated because of this. Now, since models are cached, the rotations can be stored in models that won't get re-created.

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://amblelabs.github.io/ait-wiki/guidelines).
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

None

**Changelog**

:cl:
- perf: greatly improved exterior and console rendering performance

PR #2101, opened 2026-06-12
fix: re-create exterior model in snow globe and monitor

drtheodor:
## About the PR

This PR fixes a regression introduced by #2080.

## Why / Balance

As per TODO.

## Technical details

Everything else (I believe) is fine to use the cached model, but snow globe and monitor don't animate it, meaning they don't reset the bone positions. So the exterior may appear as having doors opened, etc.

## Requirements

## Breaking changes

**Changelog**


Addi3 (approved):
LGTM

PR #2104, opened 2026-06-13
fix: restart client crash

drtheodor:
## About the PR

This PR fixes a client-side crash that happens on server restart.

## Why / Balance

It's a crash.

## Technical details

Example crash log:
```log
java.lang.NullPointerException: Cannot invoke "dev.amble.ait.core.tardis.handler.SiegeHandler.isActive()" because the return value of "dev.amble.ait.client.tardis.ClientTardis.siege()" is null
	at knot//dev.amble.ait.client.renderers.doors.DoorRenderer.render(DoorRenderer.java:66)
	at knot//dev.amble.ait.client.renderers.doors.DoorRenderer.method_3569(DoorRenderer.java:35)
	at knot//net.minecraft.class_824.method_23079(class_824.java:90)
        ....
```
This issue is caused by 2 things:
- multithreading (rendering thread vs main client thread)
- `ClientTardis#dispose`
What's happening is a client tardis is getting disposed of, but the rendering thread doesn't catch up to that and tries rendering right after the tardis got disposed, making it invalid and cause an NPE.

This PR's solution is to remove the `#dispose` implementation, which will make the job harder for the GC but ideally shouldn't be too hard on the modern Java GCs.

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: clients no longer crash on server restart/kick

### DIFF
--- a/src/main/java/dev/amble/ait/client/tardis/ClientTardis.java
+++ b/src/main/java/dev/amble/ait/client/tardis/ClientTardis.java
@@ -76,14 +76,14 @@ public class ClientTardis extends Tardis implements Disposable {
 
     @Override
     public void dispose() {
-        this.desktop.dispose();
-        this.desktop = null;
+        // this.desktop.dispose();
+        // this.desktop = null;
 
-        this.exterior.dispose();
-        this.exterior = null;
+        // this.exterior.dispose();
+        // this.exterior = null;
 
-        this.handlers.dispose();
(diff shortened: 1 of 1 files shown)


Loqor (approved):
bro who even wrote that shit
LGTM

drtheodor:
i did

drtheodor:
:P

PR #2121, opened 2026-06-24
Fix ghost blocks during interior reconfiguration

Peter1169:
## About the PR

Fixes TARDIS interior reconfiguration being applied while players are still inside after the countdown.

The countdown still starts as before, but once it finishes the interior is only rewritten once the old interior is clear. Players with life support are evacuated as before. Creative and spectator players are also evacuated when life support is missing, because they can otherwise survive the reconfiguration and keep tracking the old chunks. Survival/adventure players without life support keep the lethal behaviour and are removed by reconfiguration damage.

## Why / Balance

This keeps the current design intent: no life support remains lethal for survival players, while life support saves players by moving them outside.

The missing case was players that the damage path could not remove, especially creative players. Without removing those players first, `changeInterior()` can rewrite the desktop while the client is still inside it, which causes ghost blocks.

## Technical details

- Removed the fixed delayed `changeInterior()` call after the regeneration countdown.
- While `regenerating`, the handler now evacuates or damages players and only calls `changeInterior()` once the interior is empty and the desktop is no longer changing.
- Takes a snapshot of the interior player list before teleporting, so evacuation does not mutate the iterated list.
- Teleports players outside when life support is enabled, or when the player is creative/spectator.
- Applies repeated `ait:interior_change_damage_type` damage to survival/adventure players without life support for 10 seconds.
- If the player is still alive after that, applies one final `interior_change` hit with `Float.MAX_VALUE`, then falls back to out-of-world damage and `kill()`.
- Added damage type tags for armor, resistance, and invulnerability bypass.
- The out-of-world damage and `kill()` fallback should not normally be reached in vanilla because `ait:interior_change_damage_type` uses the bypass tags, but it is kept as a compatibility path for mods that hardcode immortality or invulnerability behaviour without respecting those tags.
- Added English and Spanish death translations.

## Media

N/A

## Requirements

## Breaking changes

N/A

**Changelog**
:cl:
- fix: prevent TARDIS interior reconfiguration from rewriting the interior while creative or otherwise non-killable players are still inside

### DIFF
--- a/src/main/java/dev/amble/ait/core/tardis/handler/InteriorChangingHandler.java
+++ b/src/main/java/dev/amble/ait/core/tardis/handler/InteriorChangingHandler.java
@@ -64,6 +64,9 @@ public class InteriorChangingHandler extends KeyedTardisComponent implements Tar
     private static final BoolProperty QUEUED = new BoolProperty("queued");
     private static final BoolProperty REGENERATING = new BoolProperty("regenerating");
     private static final int MIN_FUEL_COST = 5000;
+    private static final int FORCE_KILL_TICKS = 20 * 10;
+    private static final float INTERIOR_CHANGE_DAMAGE = 1000.0F;
+    private static final float FINAL_INTERIOR_CHANGE_DAMAGE = Float.MAX_VALUE;
 
     public static final int MAX_PLASMIC_MATERIAL_AMOUNT = 8;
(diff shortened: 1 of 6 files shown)


Peter1169:
Closing this PR because the interior reconfiguration bug is now fixed on `main` by [73da543e3](https://github.com/amblelabs/ait/commit/73da543e3f7564a83017b810f24e6c7213214b8a)

PR #2127, opened 2026-07-04
Synchronize the Chinese translation of 'block.ait.untempered_schism'

smallmoss233:
Synchronize the Chinese translation of 'block.ait.untempered_schism', that's all


vmbbi (approved):
LGTM

PR #2131, opened 2026-07-06
fix: fast travel autopilot behavior

drtheodor:
## About the PR

This PR fixes fast travel on autopilot (e.g. with stat remote or sonic).

## Why / Balance

TARDIS would hang at 100% flight completion and not land.

## Technical details

Idk really, travel code complicated but most likely it sees that the target ticks are 0 and tries to reconsider going into a loop and never landing. Either way, this fix just makes it so fast travel returns target ticks as 1 instead of 0.

## Requirements

## Breaking changes

**Changelog**


Loqor (approved):
LGTM

PR #2132, opened 2026-07-06
fix: random matrix energizer crash

drtheodor:
## About the PR

This PR fixes a very weird crash that happens mid-tick under a certain very rare condition.

## Why / Balance

Crash.

## Technical details

Basically, whoever wrote this should be hanged. If `randomTick` gets called after the block was broken, the thing will crash because it will attempt to grab its own block state AGAIN (even though it has the state already bruh).

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: random matrix energizer crash


Loqor (approved):
LGTM but also this was me sorry lol

PR #2136, opened 2026-07-08
Fix Repair Time Increasing

Loqor:
## About the PR
Check if the TARDIS is landed before adding repair ticks.

## Why / Balance
Makes the repair time go up even when landed which is odd.

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: repair time kept increasing even when landed when the throttle was active


rapbattlegod32 (approved):
LGTM

PR #2141, opened 2026-07-09
Fix Sonic Port

Loqor:
## About the PR
I fixed the sonic port so right-clicking on the port when a sonic or handles is in it with a sonic or handles in your hand doesn't replace what's in the port - plus, I fixed not being able to right-click to take out the sonic/handles when the power's off. 

## Why / Balance
Because.

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: when handles/a sonic are in the sonic port, they got replaced if you were holding a sonic or handles.
- fix: you can now take out whatever from the sonic port with right click when the powers out.


Addi3 (approved):
LGTM

PR #2155, opened 2026-07-16
Immersive Portals Wouldn't Get Removed When Unloaded

Loqor:
## About the PR
Portals just wouldn't get removed if they were unloaded if the TARDIS was called with a sonic/stattenheim remote.

## Why / Balance
Leaves portals all over the place

## Requirements

## Breaking changes

**Changelog**


drtheodor:
so one way to fix this is:
1. make `TardisPortal`s retain the tardis they're linked to (can be done WITHOUT tardis refs, as this is purely serverside info)
2. on `TardisPortal#readNbt` check if the TARDIS its linked to is real and check whether the entity's id matches with the appropriate entity ref
3a. if not, `#discard`
3b. if yes, do nothing

drtheodor (approved):
LGTM

PR #2156, opened 2026-07-16
Fix incorrect Astral Map distance (measured from interior player position)

### LINKED ISSUE #2029: Incorrect Calculation on structure distance
### What happened?

Hi,

the number of blocks shown when using the Astral Map is incorrect
1. Go to the astral map
2. Click on anything tbh, im using a mod called "waysigns in this example"
3. wait for the tardis to tell you if the structure is found or no and check the block distance

Image

As you can see from the image my current X Coord is -9616 (the found block is -9952 a difference of +/-336)
my current z coord is -4146 (the found block is -3652, a difference of +/-194)

336 + 194 = 530 blocks (+/- whatever the z coord is)

output is 10,741 blocks away which isn't correct

### Version

1.2.12

### Mod Loader

fabric

### Build number & branch

_No response_

### Logs & screenshots

_No response_

Nitjsefnie:
## About the PR

The Astral Map's structure and biome finders reported wildly wrong "blocks away" numbers — the issue shows ~10,741 for a structure ~600 blocks away, and a second case of ~12,000 where ~1,658 was expected (the non-constant ratio ruled out a simple unit error).

I couldn't reproduce it live (no client in my build environment), so this is a reasoned fix against the coordinate model plus the reporter's numbers rather than an observed in-game before/after — worth a maintainer sanity-check in-game before merge.

*Disclosure: this contribution was made with AI assistance (Claude), reviewed and built by a human-directed process before submission.*

## Why / Balance
Fixes #2029.

## Technical details

Root cause is a coordinate-space mix-up. Both `handleStructureRequest` and `handleBiomeRequest` in `AstralMapBlock.java` only run when the player is inside the TARDIS interior dimension (they're guarded by `TardisServerWorld.isTardisDimension(world)`). The search runs in the *target* world from the TARDIS's exterior position (`tPos.getPos()` / `currentPos.getPos()`), and the located structure/biome is a coordinate in that target world — but the distance was computed against `player.getPos()`, which is the player's position in the *interior* dimension, an unrelated coordinate space. Subtracting interior coords from target-world coords produces the garbage distance.

The fix measures the distance from the TARDIS's exterior search origin (the same reference the result was located from and reported relative to), in both finders:

- structure: `newPos.getSquaredDistance(player.getPos())` → `newPos.getSquaredDistance(tPos.getPos())`
- biome: `locatedBiome.getSquaredDistance(player.getPos())` → `locatedBiome.getSquaredDistance(start)`

(`tPos.getPos()` rather than the local `pos` because `pos` is reassigned and so isn't capturable by the async `thenOnServerThread` lambda; it holds the identical value.)

## Requirements

## Breaking changes

None

**Changelog**

:cl:
- fix: incorrect astral map distance calculation

### DIFF
--- a/src/main/java/dev/amble/ait/core/blocks/AstralMapBlock.java
+++ b/src/main/java/dev/amble/ait/core/blocks/AstralMapBlock.java
@@ -141,7 +141,7 @@ public class AstralMapBlock extends BlockWithEntity implements BlockEntityProvid
                 if (newPos != null) {
                     player.sendMessage(Text.translatable(
                             "block.ait.astral_map.finder.found", newPos.getX(), newPos.getY(), newPos.getZ(),
-                            Math.round(Math.sqrt(newPos.getSquaredDistance(player.getPos())))), false);
+                            Math.round(Math.sqrt(newPos.getSquaredDistance(tPos.getPos())))), false);
                     tardis.travel().destination(destination -> destination.pos(newPos));
                 } else {
                     player.sendMessage(Text.translatable("block.ait.astral_map.finder.structure_not_found"), false);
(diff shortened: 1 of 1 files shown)


drtheodor (approved):
LGTM

PR #2166, opened 2026-07-30
Expand TARDIS home locations

Peter1169:
## About the PR

This expands the existing TARDIS home location without including the automatic recovery protocols, home-bound subsystems, Biodata Restoration or TARDIS Temperament behavior from the original version of this PR.

- Sneak-using the telepathic circuit with an empty hand sets the configured home as the current destination.
- Handles accepts `handles go home`, which selects the same destination and starts a normal flight.
- Using a coral fragment on the telepathic circuit while landed can reassign the TARDIS home, subject to the existing loyalty requirement and a configurable 60-minute cooldown.
- Exact home positions are globally exclusive by dimension and block position. Facing does not make two otherwise identical positions different.
  - Claims include loaded, dormant and legacy TARDIS files.
  - TARDIS creation, the home command and coral-fragment reassignment reject occupied or unavailable positions.
  - A planted TARDIS coral at an occupied home remains at its initial growth stage and warns nearby players until it is moved or the position becomes available.

The rest of the original implementation is being separated into focused follow-up PRs.

## Why / Balance

The home position already existed but had little direct use after a TARDIS was created. This makes it a reliable destination while keeping travel fully manual and subject to the normal flight flow, fuel cost, speed and dimension restrictions.

Unique exact-home claims prevent two TARDISes or corals from occupying the same block and avoid ambiguity for later home-based systems. The relocation cooldown prevents repeatedly moving a home to follow the player or bypass location-based restrictions.

## Technical details

- Added a server-side exact-home claim index covering loaded, dormant and legacy TARDIS data.
- Claims compare dimension and `BlockPos`; rotation remains part of the stored home destination but not claim identity.
- Claim validation is applied consistently during TARDIS creation, loading, removal, administrative home changes and coral-fragment reassignment.
- Persisted TARDIS files are inspected incrementally and fail closed while a claim cannot be verified safely.
- Added shared home-position comparison helpers through `TardisHomeUtil`.
- Added generated `en_us` and `es_es` text for the new interactions and validation results.

## Media

N/A

## Requirements

## Breaking changes

N/A

**Changelog**
:cl:
- feat: The telepathic circuit and Handles can now target a TARDIS's configured home.
- feat: Exact TARDIS home positions are unique, including homes belonging to dormant TARDISes.
- feat: A TARDIS home can be reassigned with a coral fragment while landed, subject to a configurable cooldown.


rapbattlegod32:
ait 1.3 genuinely never seeing the day of light

src/main/java/dev/amble/ait/client/tardis/manager/ClientTardisManager.java
```diff
@@ -179,6 +179,7 @@ public void getTardis(MinecraftClient client, UUID uuid, Consumer<ClientTardis>
 
     @Override
     public void reset() {
+        ClientSoundManager.stopAll();
```
drtheodor: that should be handled by the individual sounds not ClientTardisManager

src/main/java/dev/amble/ait/client/screens/MonitorScreen.java
```diff
@@ -351,7 +355,7 @@ protected void drawTardisExterior(DrawContext context, int x, int y, float scale
 
         boolean isHorriblyUnscaled = variant.equals(ClientExteriorVariantRegistry.DOOM);
 
-        boolean isExtUnlocked = tardis.isUnlocked(variant.parent());
+        boolean isExtUnlocked = DeathRetconClient.isUnlocked(tardis, variant.parent());
```
drtheodor: why was `#isUnlocked` method moved from `Tardis` to some unrelated class?

src/main/java/dev/amble/ait/compat/gravity/GravityHandler.java
drtheodor, line 1: rude stuff should be a part of a separate PR

src/main/java/dev/amble/ait/compat/portal/PortalsHandler.java
```diff
@@ -62,6 +61,11 @@ public static void init() {
             handler.generatePortals();
         });
 
+        TardisEvents.DOOR_CLOSE.register((tdis) -> {
```
drtheodor: we don't do that because the door could still be considered "open" for the player. 
it'd make more sense to let players exit while the door is closing (previous behavior) but ig that's up to discussion.

drtheodor: also, again, shouldn't be a part of this PR

drtheodor (changes requested):
just a simple shallow review because of how big this PR is, I'm seeing some stuff that could potentially be separate PRs which should help with actually reviewing it (15kloc of changes...)

E.g.:
- death retcon (which should ideally be renamed to something else, at least in the translation files)
- "rude" mode
- etc (haven't read the entire changelog yet)
Could be all separate PRs.

Peter1169:
Thanks, agreed. I have reduced this PR to the home-location feature itself and split the rest into focused PRs:

- #2167 adds the automatic homeward and recovery protocols.
- #2168 moves client sound cleanup into the individual sound handlers.
- #2169 adds the home-bound systems and the renamed Biodata Restoration feature.
- #2170 adds Temporal Recovery independently.
- #2171 adds the renamed TARDIS Temperament system, including the directional Immersive Portals behavior.

#2166 now only covers configuring and manually targeting a TARDIS home, exact-position ownership and home reassignment. Its title, description and changelog have been updated to match that reduced scope.

The former Death Retcon name has been replaced throughout by Biodata Restoration, the former Rude TARDIS name has been replaced by TARDIS Temperament, and the punitive Nether Star consequence is now Timeline Erasure. The corresponding English and Spanish translations, code identifiers and configuration names were updated together.

src/main/java/dev/amble/ait/client/tardis/manager/ClientTardisManager.java
Peter1169: Moved this out to #2168. The individual flight music, interior hum and sonic handlers now validate their own client context and stop or reset the state they own. `ClientTardisManager` no longer has to maintain a central list of sounds to shut down.

src/main/java/dev/amble/ait/client/screens/MonitorScreen.java
Peter1169: Restored this responsibility to `Tardis#isUnlocked` in #2169. `MonitorScreen` continues to call `tardis.isUnlocked(variant.parent())`; Biodata Restoration does not own or replace the general unlock query.

src/main/java/dev/amble/ait/compat/gravity/GravityHandler.java
Peter1169: Removed all of this from #2166. It is now isolated in #2171 under the renamed TARDIS Temperament feature, with matching class, component, configuration and translation names.

src/main/java/dev/amble/ait/compat/portal/PortalsHandler.java
Peter1169: Moved this to #2171 and kept the previous outward behavior. The portal is no longer made globally non-teleportable when closing. Only an attempted entry through the exterior portal is checked for the affected REJECT player, while the interior portal remains available for exiting for as long as the door is still considered open.

PR #2168, opened 2026-07-31
Fix TARDIS client sound lifecycle

Peter1169:
## About the PR

This makes the individual TARDIS client sound handlers stop and reset their own sounds when the client context they depend on is no longer valid.

- The current client TARDIS reference is cleared when changing to a world without a TARDIS and validated against the active client world before it is returned.
- Flight music stops itself when there is no longer a current TARDIS.
- The interior hum handler owns its complete reset and reinitialization lifecycle.
- The sonic sound handler owns its world-change cleanup.

This intentionally does not add a central sound shutdown call to `ClientTardisManager`.

## Why / Balance

This has no gameplay balance impact. It prevents TARDIS-specific sounds from continuing after leaving an interior, changing worlds, dying, disconnecting or otherwise losing the client-side TARDIS context that created them.

Keeping cleanup inside each handler also means future sound handlers can define the state they own instead of depending on an unrelated TARDIS data manager to maintain a complete global list.

## Technical details

- `ClientTardisUtil#getCurrentTardis` now verifies that its reference matches the TARDIS UUID attached to the active client world.
- World changes without a TARDIS clear the stale reference instead of leaving it available to later sound ticks.
- `ClientFlightMusicHandler`, `ClientHumHandler` and `SonicSoundHandler` perform their own context checks and cleanup.
- Hum reset stops active instances, clears the selected hum and marks the handler for safe reinitialization when a valid TARDIS becomes available again.
- The changes have passed Spotless, datagen and the complete Gradle build.

## Media

N/A

## Requirements

## Breaking changes

N/A

**Changelog**
:cl:
- fix: TARDIS hums, flight music and sonic sounds no longer continue after their client context is gone.


src/main/java/dev/amble/ait/client/util/ClientTardisUtil.java
```diff
@@ -123,14 +124,17 @@ public static void toggleAntigravs(UUID uuid) {
     }
 
     public static boolean isPlayerInATardis() {
-        return currentTardis != null;
+        return getCurrentTardis() != null;
```
drtheodor: revert this, as even though this gives a more proper result, this method (isPlayerInATardis) gets called in rendering and other high-intensity situations, where speed is of the matter and we can not just let the thread stall
we don't care for such exact timings either

src/main/java/dev/amble/ait/client/util/ClientTardisUtil.java
```diff
@@ -123,14 +124,17 @@ public static void toggleAntigravs(UUID uuid) {
...
 
     /**
      * Gets the tardis the player is currently inside
      */
     public static ClientTardis getCurrentTardis() {
-        if (currentTardis == null)
+        MinecraftClient client = MinecraftClient.getInstance();
+        UUID worldTardis = TardisServerWorld.getTardisId(client.world);
+        if (currentTardis == null || worldTardis == null
+                || !worldTardis.equals(currentTardis.getId()))
```
drtheodor: not needed (see my previous comment)

drtheodor (changes requested):
CHANGE_WORLD is already working properly, there's no need to re-query the current tardis from the cache each time (not to mention that doing string -> uuid is slightly expensive.

this method gets called in rendering a lot, and I don't see a reason to why it should re-query every call

PR #2172, opened 2026-08-01
Fix Artron transfer, fluid network and Rift energy inconsistencies

Peter1169:
## About the PR

This fixes a group of related inconsistencies across Artron storage, TARDIS refuelling, fluid networks, Rift Chunks and Rift-related devices.

- Artron holders now use consistent insertion, extraction and remainder semantics.
- Invalid, negative and non-finite fuel values are rejected or normalized safely.
- Partial transfers no longer silently lose or create Artron.
- Artron Collectors preserve energy when transferring between Rift Chunks, items and connected TARDIS engines.
- Zeiton Cluster conversion consumes exactly one cluster instead of replacing the complete stack.
- Power Converters can now recharge connected empty sources and only consume items whose complete Artron value can be accepted.
- Fluid networks distinguish between being connected and being powered.
- Network assignments are restored when chunks containing fluid links are loaded.
- Network discovery is bounded and transactional, preventing both partial assignments and unbounded scans.
- Engine Block placement validates and restores its automatically placed cables and barriers transactionally.
- Rift Chunk storage, extraction and regeneration now work correctly in fully loaded chunks.
- TARDIS Rift refuelling consumes the same amount of stored Rift Artron that it adds as a bonus.
- Rift Scanner searches cover the complete surrounding area, preserve the target dimension and no longer force candidate chunks to load.
- Rift rewards verify sonic capacity and safely handle empty reward tags.
- Untempered Schism creation is validated and completed transactionally.
- Added server configuration for Rift Chunk capacity, regeneration and TARDIS refuelling rates.

## Why / Balance

Artron is intended to behave as a conserved resource when it moves between storage systems. Previously, different holders used incompatible transfer semantics, which could cause excess energy to disappear, partially accepted items to be consumed, or a smaller Rift Chunk extraction to produce a larger TARDIS refuelling bonus.

The default values preserve the intended existing refuelling rates:

- Rift Chunks generate with between 2000 and 4000 AU.
- Rift Chunks regenerate 1 AU per second.
- Normal landed refuelling generates 140 AU per second.
- Refuelling inside a Rift Chunk can provide an additional 40 AU per second.

The Rift bonus now consumes the corresponding amount from the Rift Chunk instead of multiplying a smaller extraction. Ambient refuelling remains independent because it represents environmental generation rather than stored Rift energy.

Power Converters and Zeiton Clusters only consume complete items when their full Artron value can be accepted. This prevents resources from being partially destroyed when a destination is almost full.

Fluid networks are limited to 4096 loaded nodes per rebuild. Oversized networks retain their previous complete assignments rather than receiving a partial assignment or causing an unbounded synchronous traversal.

The new configuration options allow modpack and server owners to adjust these rates while preserving the existing intended balance by default.

## Technical details

- Standardized the shared Artron transfer contracts:
  - `addFuel` returns the unaccepted remainder.
  - `insertFuel` returns the amount accepted.
  - `extractFuel` returns the amount extracted.
  - Existing `removeFuel` compatibility wrappers remain available.
- Existing methods and item NBT keys are preserved.
- The new interface helpers use default implementations, so existing implementations are not required to add new methods.
- Existing callers which ignore transfer return values remain source-compatible.
- Added clamping and normalization for stored fuel and legacy item NBT.
- Added equivalent bounded insertion and extraction helpers to `IFluidSource`.
- Preserved compatibility with legacy fluid sources whose setters already invoke zero-crossing callbacks.
- Added `IFluidSource#setLevelNotifies` so legacy notifying setters and silent setters do not emit duplicate network events.
- Updated Artron Collector item and block storage to use the shared transfer contracts.
- Corrected the collector tick interval and default item-stack initialization.
- Collector transfers now preflight source and destination capacity and restore unexpected remainders.
- Zeiton Cluster conversion supports both hands and safely inserts or drops the resulting Charged Zeiton Crystal.
- Console Zeiton insertion uses the combined available capacity of the primary tank and enabled Emergency Power reserve.
- Removed the unreachable Handles charging loop while retaining the existing handler interfaces for addon compatibility.
- Power Converter player and hopper interactions now share the same whole-item transfer calculation.
- Fluid networks now:
  - Rebuild when chunks containing fluid links load.
  - Avoid loading neighbouring chunks during discovery.
  - Select sources deterministically.
  - Prefer powered sources with more stored energy.
  - Retain empty sources as valid connections so they can be recharged.
  - Rebuild when a source crosses between empty and powered states.
  - Apply assignments only after the complete loaded component has been discovered.
  - Abort safely at the 4096-node limit.
  - Retain their previous assignments after an oversized rebuild is rejected.
  - Rate-limit oversized-network warnings.
  - Avoid rediscovering the same oversized component from every remaining node in a loaded chunk.
- Engine Block placement performs a complete preflight and rolls back only blocks changed by that placement.
- Engine cleanup removes only the automatically created horizontal connections and barriers.
- Cable placement validation accepts connection and waterlogging state updates.
- The TARDIS desktop now tracks the Engine Block position used for network rebuilding.
- Rift Chunk data now uses separate persistent current and maximum Artron attachments.
- Existing attachment data is migrated lazily without deleting stored energy.
- Rift Chunk operations use already loaded `Chunk` instances instead of requiring `ProtoChunk`.
- Partial extraction reports the actual amount removed.
- Insertion returns any amount which did not fit.
- Regeneration is clamped to the configured maximum.
- Rift Chunk queries do not force chunks to load or generate.
- TARDIS refuelling separately accounts for ambient generation and the Rift Chunk bonus.
- Rift energy extraction is limited to the total capacity available across the enabled reserve and primary tank.
- Unexpected Rift transfer remainders are returned to the source chunk.
- Refueler controls only report success when the refuelling state actually changes.
- Rift Scanner targets now include dimension, chunk X and chunk Z.
- Chunk `(0, 0)` is supported as a normal target instead of representing an absent target.
- Scanner searches use complete square rings and clear stale targets after failed searches.
- The scanner only inspects stored energy when a candidate chunk is already loaded.
- The normal 250 AU scanner threshold is clamped to the configured Rift Chunk capacity.
- Zero-capacity Rift Chunks are not selected as useful targets.
- Rift consumption requires enough sonic capacity for the complete reward.
- Empty Rift reward tags no longer produce invalid random bounds.
- Rift terrain propagation resolves the correct heightmap across chunk borders and preserves relevant slab, stair and waterlogging properties.
- Untempered Schisms validate their Rift Chunk before consuming the placement item.
- Schism Rift creation only commits after entity spawning, fuel consumption and block-state changes all succeed.
- Failed Schism creation restores both the entity and stored energy state.
- Created Rift references are stored by UUID and cleaned safely when the entity is missing.
- Added server configuration options for:
  - Minimum Rift Chunk Artron capacity.
  - Maximum Rift Chunk Artron capacity.
  - Rift Chunk regeneration per second.
  - Ambient TARDIS refuelling per second.
  - Rift Chunk refuelling bonus per second.
- Invalid, negative or reversed configuration values are normalized safely.
- Added generated configuration translations and descriptions.
- Added lightweight JUnit coverage for:
  - Accepted amounts and remainders.
  - Partial and complete extraction.
  - Fuel conservation at storage boundaries.
  - Invalid and non-finite transfer values.
  - Legacy item NBT normalization.
  - Fluid-source zero-crossing notifications.
  - Protection against duplicate gain and loss callbacks.
  - Configuration bounds and default values.
  - Scanner thresholds below 250 AU.
- The tests are build-time only and do not add a runtime dependency.

## Media

N/A

## Requirements

## Breaking changes

N/A

**Changelog**
:cl:
- fix: Artron transfers now preserve accepted, extracted and remaining energy across items, TARDIS fuel storage and connected machines.
- fix: Rift Chunks now store, regenerate and provide Artron consistently without forcing candidate chunks to load or generate.
- fix: Artron Collectors, Power Converters, fluid networks, Rift Scanners and Untempered Schisms now handle empty, full and chunk-reload states correctly.
- tweak: Added server configuration for Rift Chunk capacity and regeneration and for ambient and Rift-assisted TARDIS refuelling rates.


src/main/java/dev/amble/ait/core/blockentities/EngineBlockEntity.java
```diff
@@ -26,7 +30,16 @@
...
+    private static final Direction[] HORIZONTAL_DIRECTIONS = {
+            Direction.NORTH, Direction.SOUTH, Direction.WEST, Direction.EAST
+    };
+
     private boolean firstTickHandled;
+    private boolean suppressFillCleanup;
+
+    private record FillBlock(BlockPos pos, BlockState state) {}
+
+    private record ReplacedBlock(BlockPos pos, BlockState state) {}
```
drtheodor: those can be the same record
also, instead of using a List<FillBlock>/List<ReplacedBlock> it can use a Map<BlockPos, BlockState>

src/main/java/dev/amble/ait/core/blockentities/UntemperedSchismBlockEntity.java
```diff
@@ -74,73 +85,96 @@ public double getCurrentFuel() {
...
     }
 
     @Override
     public void tick(World world, BlockPos pos, BlockState state, UntemperedSchismBlockEntity blockEntity) {
         if (!(world instanceof ServerWorld serverWorld))
             return;
 
-        if (this.hasCreatedRift)
+        if (!UntemperedSchismBlock.canCreateAt(serverWorld, pos)) {
+            serverWorld.setBlockState(pos, Blocks.LODESTONE.getDefaultState(), Block.NOTIFY_ALL);
```
drtheodor: uh... why?

src/main/java/dev/amble/ait/core/blocks/UntemperedSchismBlock.java
```diff
@@ -84,28 +84,26 @@ public VoxelShape getCullingShape(BlockState state, BlockView world, BlockPos po
...
-        ChunkPos chunkPos = new ChunkPos(pos);
-
-        if (!isConsumable(manager, chunkPos)) {
+        if (!canCreateAt(serverWorld, pos)) {
             if (placer != null) {
                 placer.sendMessage(Text.translatable("message.ait.riftscanner.info3"));
             }
             world.playSound(null, pos, SoundEvents.BLOCK_BEACON_DEACTIVATE,
                     SoundCategory.BLOCKS, 1, 0.5f);
+            world.setBlockState(pos, Blocks.LODESTONE.getDefaultState(), Block.NOTIFY_ALL);
```
drtheodor: why does it place a lodestone?

drtheodor (changes requested):

src/main/java/dev/amble/ait/core/blocks/UntemperedSchismBlock.java
Peter1169: It was meant as a visible fallback when the Schism was placed outside a Rift Chunk, but replacing it with a lodestone was the wrong behavior and could still consume the Schism item. I removed it; invalid placement is now rejected before the item is consumed

src/main/java/dev/amble/ait/core/blockentities/UntemperedSchismBlockEntity.java
Peter1169: This was the same fallback for Schisms that somehow already existed outside a Rift Chunk. I removed the conversion here too. The tick check now leaves an invalid Schism inert instead of replacing it with an unrelated block

PR #2186, opened 2026-08-27
perf(render): cut duplicate draws and cache the animation bone lookup

duzos:
## About the PR

Client FPS work, driven by measurement rather than inspection.

Eleven render optimisations, four correctness fixes found along the way, and the
profiling harness that found them. Every optimisation is paired with a per-frame
counter that proves the mechanism changed, not just that a number got smaller.
Several early hypotheses about where the cost lay turned out to be wrong, and
those counters are what caught them.

## Scenario results

Every scenario below was run on `origin/main` and on this branch, same machine,
same session, same rcon-driven scenario files, 10s vanilla client profiler dumps.

Frame times are absolute milliseconds from the dump's own time span, not the
relative percentages the file prints.

| Scenario | main ms | main fps | this branch ms | this branch fps | fps |
|---|---|---|---|---|---|
| Empty world | 1.55 | 645.6 | 1.39 | 722.0 | +12% |
| 1 exterior, doors shut | 1.59 | 629.8 | 1.86 | 536.8 | -15% |
| 1 exterior, portal open | 2.21 | 452.6 | 1.50 | 666.3 | +47% |
| 5 exteriors | 4.50 | 222.4 | 3.38 | 295.5 | +33% |
| 20 exteriors | 11.27 | 88.7 | 7.11 | 140.6 | +58% |
| 40 exteriors | 20.81 | 48.0 | 11.95 | 83.7 | +74% |
| 20 exteriors, 110 blocks out | 10.86 | 92.1 | 7.11 | 140.7 | +53% |
| 20 exteriors behind the camera | 7.20 | 138.9 | 1.40 | **715.3** | +415% |
| 1 booth exterior | 3.31 | 302.0 | 2.17 | 461.4 | +53% |
| 5 booth exteriors | 5.14 | 194.7 | 3.65 | 274.3 | +41% |
| 20 booth exteriors | 14.28 | 70.0 | 8.64 | 115.8 | +65% |
| 40 booth exteriors | 25.88 | 38.6 | 14.67 | 68.2 | +76% |
| Doom exterior | 3.89 | 257.4 | 3.00 | 333.2 | +29% |
| Siege mode | 1.54 | 647.4 | 1.61 | 621.0 | -4% |
| Shields up | 4.61 | 217.1 | 3.39 | 295.4 | +36% |
| Alarm on | 4.22 | 236.7 | 3.21 | 311.4 | +32% |
| Unpowered | 4.61 | 217.1 | 3.17 | 315.2 | +45% |
[...]

PR #2191, opened 2026-09-08
fix(renderers): always sort the emissive layer's quads

duzos:
## About the PR

Fixes the rendering regression reported against beta 2816: console monitor
animations missing on Renaissance and Toyota, and emissions showing through the
panel around the power switch.

Reverts the one change in #2186 that caused it, and keeps the reasoning in the
code so it does not get dropped again for the same wrong reason.

## Why / Balance

No balance impact. This restores the pre-#2186 look of every emissive draw.

Costs back the roughly 0.6 ms the sort was measured to save on Copper's
emission pass. On the numbers from #2186 that is about 117 fps to about 109 in
the Copper interior scenario, against 39 fps on the pre-#2186 baseline, so the
bulk of that PR's gain is unaffected.

## Technical details

`AITRenderLayers.emissive` builds the layer with:

```
.cull(DISABLE_CULLING)
.transparency(TRANSLUCENT_TRANSPARENCY)
.writeMaskState(COLOR_MASK)          // depth TEST, no depth WRITE
.depthTest(LEQUAL_DEPTH_TEST)
```

#2186 dropped the `translucent` flag on this layer, which is what hands
`BufferBuilder` a sorter, to save the per-quad primitive centres and the
explicit index buffer.

The justification given was that 106 of the mod's 113 emission textures are
strictly binary alpha, where `SRC_ALPHA` blending is order-independent, so only
textures with real partial alpha needed the sorted layer.

That is wrong in one specific way. **Binary alpha makes the blend
order-independent, not the occlusion.** Because the layer writes no depth,
nothing in the batch occludes anything else in it, and because back face culling
is disabled, both faces of every part are submitted. Whichever quad is drawn
last simply wins. So submission order decides what ends up on top regardless of
alpha: a part's own back face, or a light sitting behind a panel, paints over
the front of it.

The same mechanism was already written down correctly in this file, in the
javadoc for the sorted accessor, about the eye of harmony star: "with culling
disabled so both faces of both shells are submitted; unsorted, the opaque core
would land last and paint over the shell that is supposed to veil it." It simply
was not applied to the console layer.

Symptom mapping:

- **Emissions through the panel** is the direct case. `powerlights2` is one of
  Toyota's animated bones, and it is behind the panel it now paints over.
- **Renaissance and Toyota monitor animations** are the same cause. What
  animates on those monitors is emissive: Renaissance keyframes `symbol`, a
  child of `monitor`, plus `lamp1` through `lamp6`, `rotorlight` and three
  blinkers; Toyota keyframes `gallifreyan2`, `gallifreyan3`, `uppertimepiece`,
  `powerlights2` and the rest. Those are small moving glow quads, and with the
  ordering wrong they are covered and stop reading as animated.
- Renaissance and Toyota, specifically, because they are the only console models
  that both carry a monitor and are affected in the overlapping-glow way this
  produces. Hudolin shares their root transform but has no monitor at all.

With unsorted gone there is only one emissive layer, so the two accessors are
merged. That also removes the extra flush the sorted accessor's own javadoc
described, where an exterior emission texture reached both layers in one frame
and `Immediate` keys buffers on layer identity.

## Not fixed here

The third item in the report, the Copper monitor text looking wrong, is a
separate change (`ed57ab87c`, monitor text from `drawWithOutline` to a shadowed
draw on `POLYGON_OFFSET`). Copper is the only model that was converted, which is
why only its text differs. It is cosmetic and wants a look at the intended
appearance rather than a blind revert, so it is left out of this PR.

## Media

Needs a before and after on a Renaissance or Toyota console with the power on.

## Requirements

None.

## Breaking changes

`AITRenderLayers.tardisEmissiveCullZOffsetSorted(Identifier)` is removed. There
is one emissive layer now and `tardisEmissiveCullZOffset(Identifier)` returns it,
which is what the three callers were updated to.

**Changelog**

:cl:
- fix: console monitor animations are visible again on Renaissance and Toyota
- fix: emissive lights no longer draw through the panels in front of them


drtheodor (approved):
LGTM

